PR #98, opened 2021-09-29
WIP: feat: initial implementation of microvm create

### LINKED ISSUE #31: Implement initial vmspec create plan (measure, plan, act)
Basic Firecracker API calls

### LINKED ISSUE #71: Microvm repo check if spec is the same
When saving the microvm spec in the repo we need to check if there is actually an update to the spec or not. If nothing has changed, then we don't need to take any action.

If we currently try saving where there is no update then an error is returned.

Easiest way is to SHA256 hash the content and compare it to the digest of the content in the containerd content store.

richardcase:

This change contains the initial implementation of microvm create.

Fixes #31 
Fixes #71

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Create microvm via API call has been implemented.
```


richardcase:
> This PR is so big! Please, split it blush

Even GitHub Actions are telling me i've done a bad thing!

api/types/microvm.proto
```diff
@@ -53,17 +57,37 @@ message Kernel {
...
+    TAP = 1;
+  }
+  // GuestDeviceName is the name of the network interface to create in the microvm.
+  string guest_device_name = 1;
+  // IfaceType specifies the type of network interface to create for use by the guest.
+  IfaceType type = 2;
   // AllowMetadataReq indicates if the network interface allows metadata requests.
-  bool allow_metadata_req = 1;
+  bool allow_metadata_req = 3;
   // GuestMAC allows the specifying of a specifi MAC address to use for the interface. If
```
yitsushi, line 86:
```suggestion
  // GuestMAC allows the specifying of a specific MAC address to use for the interface.
```

api/types/microvm.proto
```diff
@@ -53,17 +57,37 @@ message Kernel {
...
-  // not supplied a autogenerated MAC address will be used.y 
-  optional string guest_mac = 2;
-  // GuestDeviceName is the name of the network interface to create in the microvm. If this
-  // is not supplied than a device name will be assigned automatically
-  optional string guest_device_name = 3;
+  // not supplied a autogenerated MAC address will be used.
```
yitsushi, line 87:
```suggestion
  // If not supplied an auto-generated MAC address will be used.
```

core/ports/repositories.go
```diff
@@ -14,7 +14,8 @@ type MicroVMRepository interface {
 	Delete(ctx context.Context, microvm *models.MicroVM) error
 	// Get will get the microvm spec with the given name/namespace.
 	Get(ctx context.Context, name, namespace string) (*models.MicroVM, error)
-	// GetAll will get a list of microvm details.
+	// GetAll will get a list of microvm details. If namespace is an empty string all
+	// details of microvms will be returned.
 	GetAll(ctx context.Context, namespace string) ([]*models.MicroVM, error)
```
yitsushi, line 19: I know it's already there, but wouldn't be `List` a better name for this functionality?

core/ports/services.go
```diff
@@ -40,21 +61,71 @@ type EventEnvelope struct {
...
 type ImageService interface {
-	// Get will get (i.e. pull) the image for a specific owner.
-	Get(ctx context.Context, input GetImageInput) error
-	// GetAndMount will get (i.e. pull) the image for a specific owner and then
+	// Pull will get (i.e. pull) the image for a specific owner.
+	Pull(ctx context.Context, input *ImageSpec) error
+	// PullAndMount will get (i.e. pull) the image for a specific owner and then
 	// make it available via a mount point.
-	GetAndMount(ctx context.Context, input GetImageInput) ([]models.Mount, error)
+	PullAndMount(ctx context.Context, input *ImageMountSpec) ([]models.Mount, error)
```
yitsushi, line 68: If it has a separate Pull, maybe a Mount function can be useful. For example if we want to control the full workflow and call `Pull`, verify, and then `Mount`. With this, we can call `Pull`, verify, and `PullAndMerge`, which is fine if it's guaranteed it will be the same and does not try to pull and/or check if the remote has been change again

yitsushi (commented):
I could manage around half way through the PR, looks good, only a few questions/suggestions.

yitsushi (approved):
Looks good to me, I suggest to merge it as it is (assuming everything is working) and create follow-up tickets/PRs. Large PRs tend to grow bigger and bigger in the name of "fix/add just this one" and it becomes harder and harder to review and test.

PR #138, opened 2021-10-13
ci: Use correct format for terraform action

### LINKED ISSUE #127: GitHub Actions fails because something is not good with the uses attribute [nightly_e2e]
**What happened:**
```
Error : .github#L1
the `uses' attribute must be a path, a Docker image, or owner/repo@ref
```

Run tests.

**How to reproduce it:**
https://github.com/weaveworks/reignite/actions/runs/1328442985

Something is off on the main branch. The PR checker works.

Callisto13:

Fixes the e2e nightlies, it appears they never actually had a successful run.

This line looks like the most likely culprit.

I suspect the intent was to use [this action](https://github.com/hashicorp/setup-terraform) @richardcase, but a typo got in the way?

Fixes #127

### DIFF
--- a/.github/workflows/nightly_e2e.yml
+++ b/.github/workflows/nightly_e2e.yml
@@ -1,13 +1,13 @@
 name: nightly_e2e
 
-on:
-  schedule:
-    - cron: 0 0 * * *
+on: [pull_request]
 
 jobs:
   e2e:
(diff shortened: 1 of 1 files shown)


Callisto13:
no merging yet as i want to run some tests

PR #144, opened 2021-10-15
Delete MicroVM

### LINKED ISSUE #65: Implement initial vmspec deletion plan (measure, plan, act
A "plan" is needed with associated steps to delete a microvm from the spec.

### LINKED ISSUE #103: Run detached fails with error when starting microvm
**What happened:**
When running a Firecracker microvm via ignite in detached mode (i.e. `--firecracker-detach=true`) we see the following error in the firecracker log:

```
2021-10-04T13:23:05.669252883 [ns1-mvm0:main:ERROR:src/firecracker/src/main.rs:48] Firecracker panicked at 'Cannot register the api event to the event manager.: A handler for the specified pollable 0 already exists.', src/firecracker/src/api_server_adapter.rs:55:14
```

Expectation is the microvm should start

**How to reproduce it:**
Start reignite with `--firecracker-detach=true` and then send a create mivrovm request.

**Environment:**

- reignite version:

### LINKED ISSUE #139: Run firecracker processes in "detached" mode
Right now, if the service dies or restarted, all microvms are terminated.

If the host has 7 microvms, it would be painful to recreate all of them on restart.

Restart can happen:
* Random failure (we don't like this, but can happen)
* Update the service (we like this)

### LINKED ISSUE #145: Create endpoint returns an error for new VMs in the same namespace
**What happened:**
After creating `ns1/vm1`, all new VMs in the same namespace like `ns1/vm2` says VM already exists.

I can create multiple VMs in the same namespace.

**How to reproduce it:**
1. Create a VM: `ns1/vm1`
2. Create a VM: `ns1/vm2`

yitsushi:

### Delete MicroVM

Reconciliation loop makes the decision if we want to use the Create,
Update, or Delete plan.

Create plan is used only if State is pending. If the state ever moved
from pending to created, we consider changes as an update, because it's
created even if only partially, the update plan should remove or add
resources.

Delete plan is used only if DeletedAt is not zero.

If it was not a Create or a Delete plan, it's an update plan. In an
ideal case, update should return an empty step list.

Right now, Volume delete is not implemented as a step. We have to check
if it's safe or not, because more than one machine can use the same
image. They are not using the same snapshot, but the snapshot is based
on the image.

Return the number of executed steps from `Actuator.Execute`, that way we
can inspect how many steps we wanted to execute and how many steps we
executed. For now it used only to determine if we have to save or not
in `reconcile`. If we executed zero steps, we don't want to save and
create a new version easy time, because it would be an infinite update
loop.

### Fix repo.Get

We are using `store.Walk(fn, filter1, filter2)`, but `Walk` calls
`ParseAll(filters...)` and it returns an `Any` filter. Even the
documentation for the functions says:

    // ParseAll parses each filter in ss and returns a filter that will return true
    // if any filter matches the expression.
    //
    // If no filters are provided, the filter will match anything.
    func ParseAll(ss ...string) (Filter, error) {

To get an `All` filter, we can join the filter strings with `,` because
`parser.selectors` creates an `All` resource for all filters separated
by a `,`.

### Run Firecracker in detached mode

The MicroVM process does not die with reignited. Read the [commit message](https://github.com/weaveworks/reignite/pull/144/commits/dd895133866710bd43414540fb19831812440628)

### Small shell tooling

I was annoyed by switching between terminal and BloomRPC back and forth, so created a small shell script that let you send grpc requests from the comfort of you tmux. Read the [commit message](https://github.com/weaveworks/reignite/pull/144/commits/d2fe581a1198f8af37703c66164a980e7316394b)

Fixes #65
Fixes #145 
Fixes #139
Fixes #103 

No more addition if possible, follow-up issues are preferred at this point for addition. This PR is already **HUGE** and went off the road with the "issue scope" in the trunk.

**Manual testing**:

1. Start `containerd`
2. Start `reignited`
3. `./hack/scripts/send.sh -m CreateMicroVM`
4. Verify Firecracker is running
5. `./hack/scripts/send.sh -m DeleteMicroVM`
6. Verify Firecracker is not running

Optional: if you kill/restart reignited, the firecracker process should stay alive.

**Checklist**:

- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/core/ports/repositories.go
+++ b/core/ports/repositories.go
@@ -19,6 +19,6 @@ type MicroVMRepository interface {
 	GetAll(ctx context.Context, namespace string) ([]*models.MicroVM, error)
 	// Exists checks to see if the microvm spec exists in the repo.
 	Exists(ctx context.Context, name, namespace string) (bool, error)
-	// ReleaseLEase will release the supplied lease.
+	// ReleaseLease will release the supplied lease.
 	ReleaseLease(ctx context.Context, microvm *models.MicroVM) error
 }
--- a/infrastructure/containerd/repo.go
+++ b/infrastructure/containerd/repo.go
@@ -171,7 +171,7 @@ func (r *containerdRepo) GetAll(ctx context.Context, namespace string) ([]*model
 	return items, nil
 }
 
-// ReleaseLEase will release the supplied lease.
+// ReleaseLease will release the supplied lease.
 func (r *containerdRepo) ReleaseLease(ctx context.Context, microvm *models.MicroVM) error {
 	mu := r.getMutex(microvm.ID.String())
 


richardcase:
> Right now, Volume delete is not implemented as a step. We have to check
if it's safe or not, because more than one machine can use the same
image. They are not using the same snapshot, but the snapshot is based
on the image.

@yitsushi  - for the time being snapshot delete is enough. However, i was thinking in the future we have our own "garbage collection" routine. In this instance perhaps its listening for the `SnapshotRemove` event which causes us to then check if there are any snapshots associated with an image....if not delete it. Or do we keep images for a period of time as a cache to make future microvms start quicker as no pull is needed and then if its not used in X days then delete.

I think keeping it around for a period of time before deleting is best.

PR #147, opened 2021-10-18
feat: add model spec validation

jmickey:

Model spec validation!

Model spec validation is provided by github.com/go-playground/validator. Validation tags are added to the model in the core/models/microvm.go file. See the validator docs for all available validation options.

The validator is added to the server as a singleton, which is initialised when NewServer is called. The validator is then called after the model spec after the request model has been converted into the internal microvm model.

Fixes #

**Checklist**:

- [x] includes documentation
- [x] adds unit tests


jmickey:
@yitsushi 

`go-playground/validator` doesn't support regex validation, will need to add this as a custom validation function, same with checking if `IsRoot=true` exists - working on these at the moment. I have addressed your other comments.

PR #154, opened 2021-10-20
Try to fix EventService test failures

### LINKED ISSUE #115: Event service integration test flake
**What happened:**
The "integration" tests for the event service aren't very stable and often have to be re-run as we get an error:

```
=== RUN   TestEventService_Integration
    event_service_test.go:29: creating subscribers
    event_service_test.go:61: subscribers waiting for events
    event_service_test.go:58: finished publishing events
    event_service_test.go:94: rpc error: code = Canceled desc = context canceled
--- FAIL: TestEventService_Integration (0.01s)
```

I expect them to pass consistently

**Environment:**

- reignite version:

yitsushi:
I was able to reproduce the issue with a helping had from cgroup.
Limiting to one CPU with a heavily quota (0.01% of my CPU) revealed the
issue.

Note before I describe what happened:
Containerd does not send messages to subscribers retrospectively, all
subscribers will receive events from the point they subscribed.

The original test created published N events and after that created
subscribers. The connection between the test and containerd is much
slower than the execution of the test, so when containrd wants to send
out the events to all subscribers, they are already there to receive
events. That's why it works on my machine and that's why it can pass on
Github Actions sometimes.

However, on a slow machine, with only one vcpu, the test and containerd
are racing for their own cpu share. In this scenario, the events are
already published before the subscribers are ready to receive them.

Solution:
Create subscribers first and then publish events.

Disclaimer:
There is a chance, all I wrote above is not entire correct, but that's
how I understand it. It does not really matter if we only check the
logic in the test. Originally it was publish->subscribe, but it's not
correct, we need subscribers before we publish events.

Fixes #115

**Checklist**:

- [x] includes documentation

### DIFF
--- a/infrastructure/containerd/event_service_test.go
+++ b/infrastructure/containerd/event_service_test.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	. "github.com/onsi/gomega"
-	"github.com/stretchr/testify/assert"
 
 	"github.com/weaveworks/reignite/api/events"
 	"github.com/weaveworks/reignite/core/ports"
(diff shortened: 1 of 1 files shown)


yitsushi:
It does not fix it XD it does a painful crash after 2 minutes.

yitsushi:
Nope still not... why

Issue #66, opened 2021-08-23
Update microvm provisioning plan to ensure its idempotent & environment aware


richardcase:
**Describe the solution you'd like:**

Change the plan/steps so that it's aware of any changes in the environment to ensure it can re-create the desired state. For example, it should be aware that the Firecracker process has dies and create/config a new one.

Bug for specified issue: #178

PR #184, opened 2021-10-29
Ability to query older versions of a vmspec

yitsushi:

With this change set, we can query older versions of a spec. We need this for update and update validation.

Related to #66

I tried to come up with a proper solution to this, but all of them had
flaws.

== Options I found and explored their possibilities

1. Add a new version argument to the Get function.

That's just ugly as hell. I like the Options pattern. Much easier to
update without breaking 300 calls in the codebase.

2. Add a new function on the Repo.

It seems odd and I think it's unnecessary because Get can handle it.

3. Create a function only on the containerd repo implementation.

That would require checks and casting everywhere we want to use it.

== Conclusion

I picked options 1, because it causes less pain not and long term.
It does not matter what content store we are using, it HAS to be able
to manage versions somehow, even if it's an external service, the
Repository implementation has to handle versions, without versions
we are are playing with a bag of venomous snakes without any kind of
antidote, maybe fun, but not safe.

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


infrastructure/containerd/repo_test.go
```diff
@@ -59,11 +72,10 @@ func TestMicroVMRepo_Integration(t *testing.T) {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(exists).To(BeFalse())
 
-	exists, err = repo.Exists(ctx, testOwnerName, testOwnerNamespace)
-	Expect(err).NotTo(HaveOccurred())
-	Expect(exists).To(BeFalse())
```
yitsushi, line 64: Removed because it  was a duplication, the previous 3 lines are exactly the same.

richardcase (approved):
I have no more comments.

Issue #146, opened 2021-10-18
[e2e]: Get initial e2e testing currently implemented endpoints


Callisto13:
E2Es should perform basic CRUD operations in a metal (Equinix) box against triggering commit sha. They should also be runnable locally.

PR #187, opened 2021-10-29
Add initial E2E tests [running locally]

Callisto13:
# The long awaited E2E tests!! Will they live up to the hype?!

Probably not since this is only a small piece from a [long list](https://github.com/weaveworks/flintlock/issues/146#issuecomment-953665693) and there is no guarantee whatsoever that they will work for anyone else.

This PR adds:
- A Dockerfile (and friends)
  - Which can be used to run a container (in a very specific way) with all the "things" one needs to play with flintlockd
  - There are some make commands to interact with said Dockerfile, I have set up a remote repo so people can pull if they don't feel like building, but that is non essential for now.
- Some initial CRUD tests for flintlockd
  - These are written in Go
  - They are all one test, because that is cheaper. This can be expanded/changed as we go.
  - They test only the happy path, as I am hoping all the control logic and edges will be caught in Units. 

This PR removes:
- All the previously existing Terraform stuff

This PR does not:
- Do many things
- One of which is run any tests in Equinix as part of a nightly build

This is just the initial bit to get us by, so much will change, for instance:
- ~Right now there are various placeholder checks in `test/e2e/e2e_test.go` which ensure a mVM has started. This was just until the Get was merged, so I am actually gonna replace that right now this second.~ Done
- The devicemapper thinpool setup is currently sitting in an `entrypoint` script and run on container start. This is very finicky when it comes to tearing down, so I am going to move that into the test itself. (Balazs it was the `trap` bit which was irritating me earlier. Like always the option you choose to be a shortcut is the most time consuming.)
- There is a very poorly named thing which handles starting and cleaning containerd and flintlockd processes (`test/e2e/utils/runner.go`), so please suggest whatever.
- The container image is built by calling a `bootstrap.sh` script. I might change it because this is kindof useful and also kindof not.
  - Useful: other people can run the script on their machine to set up their stuff so it is multipurpose
  - Not: if one bit of the script changes then basically all the image layers need to be rebuilt which is a bummer
- Does it even need to run in a container at all tho?! I mean if we have to mount `/dev` and share the `ipc` namespace ..... 
- We may also want to use more Ginkgo things for this. Eg, they have a nice teardown wrapper which will capture signals and ensure teardown/cleanup is still executed if the tests are interrupted.

Coming soon to a test suite near you:
- Configurable params!!
- A metal option!!!
- Some sexy python tooling!!
- Developer docs!!
- And much more!!!

these are all in various states of preparedness and idk which order they will appear in


jmickey (dismissed):
Brilliant stuff! Read it twice and the only thing I found was a rogue space

Callisto13:
@jmickey updated to include Get and List

test/e2e/utils/runner.go
```diff
@@ -0,0 +1,126 @@
...
+	containerdSocket = "/run/containerd-dev/containerd.sock"
+	containerdCfg    = "/etc/containerd/config-dev.toml"
+	grpcDialTarget   = "127.0.0.1:9090"
+)
+
+// Runner is a very poorly named thing and honestly idk what to call it.
+// What it does is compile flintlockd and start containerd and flintlockd.
+// So 'TestSetterUpper' did not sound as slick, but that is what it is.
+// I am happy for literally any suggestions.
+type Runner struct {
```
jmickey, line 31: Bootstrapper?

Also not coming up with any decent ideas.

jmickey (dismissed):

test/e2e/utils/runner.go
Callisto13: there is more coming to it soon, so maybe inspiration will strike

Callisto13:
sorry i had to re-push + dismiss bc apparently my last commit was not signed 🤦‍♀️

jmickey (approved):

Callisto13:
oh ffs

Callisto13:
LOL and now none of them have it

jmickey (approved):

PR #192, opened 2021-11-02
refactor: merge create and update plans and update references

### LINKED ISSUE #156: Move `plan` login into a single microvm plan
Refactor the [logic](https://github.com/weaveworks/reignite/blob/main/core/application/reconcile.go#L52) in the reconcile function that determines which plan to use so that the logic sits in the plan itself.

Roughly it will require:

- Move from 3 separate plans to 1
- Introduce 2 top level steps (Create/Update, Delete Microvm) which in turn return steps as part of their `Do`.
- If no steps are returned from the top level steps then the plan should return nothing
- Change refactor ro use the new plan

The plan will contain all the logic and this won't leak into the reconciler.

jmickey:

Merges create and update plans into a single plan builder. The update plan currently has no implementation. Update references to the plan.

The update plan code branch is reached if `vm.Spec.UpdatedAt` property is not set (defaults to 0).

Fixes #156 

**Checklist**:


core/plans/const.go
```diff
@@ -1,7 +1,8 @@
 package plans
 
 const (
-	MicroVMDeletePlanName = "microvm_delete"
-	MicroVMCreatePlanName = "microvm_create"
-	MicroVMUpdatePlanName = "microvm_update"
+	MicroVMDeletePlanName         = "microvm_delete"
+	MicroVMCreatePlanName         = "microvm_create"
+	MicroVMUpdatePlanName         = "microvm_update"
```
yitsushi, line 6: Do we need these two?

core/plans/microvm_create_update.go
```diff
@@ -54,40 +53,58 @@ func (p *microvmCreatePlan) Create(ctx context.Context) ([]planner.Procedure, er
...
 	p.ensureStatus()
-	p.steps = []planner.Procedure{}
+
+	var err error
+	switch {
+	case p.vm.Spec.UpdatedAt != 0:
+		err = p.update(ctx, ports)
+	default:
+		err = p.create(ctx, ports)
+	}
```
yitsushi, line 63: I think we can remove this completely and just walk through with all the steps, like "create microvm" and the step will say "no it's running, we don't have to do this".

@richardcase: Any idea/plan/suggestion?

core/plans/const.go
jmickey: Nope, removed

core/plans/microvm_create_update.go
jmickey: I had considered this, but I think the mental overhead of keeping them separate isn't very large, and until we actually implement the update logic it's hard to know how different the two are.

yitsushi (approved):

PR #197, opened 2021-11-03
feat: add version subcommand

### LINKED ISSUE #163: Add some way for users to discover their flintlock version
Probably can wait until we have actual releases.

I'd like to be able to run `flintlockd -v` or hit some endpoint on a running instance or look at some file or something to find out the version of flintlock I am using.

Our bug issue template asks people to provide their flintlock version, but there is no way for them to do this, unless they happen to remember the version they downloaded / sha they built from.

yitsushi:

The `version` subcommand without any flags prints out a general version
information, good to check what version we have.

    ❯ ./bin/flintlockd_amd64 version
    flintlock v0.1.0-alpha.1-11-af7a1a9

The `version` subcommand with `--short` flag prints out only the
version, useful for scripting.

    ❯ ./bin/flintlockd_amd64 version --short
    v0.1.0-alpha.1-11-af7a1a9

The `version` subcommand with `--long` flag prints out a detailed version
information, useful for bug reports.

    ❯ ./bin/flintlockd_amd64 version --long
    flintlock
      Version:    v0.1.0-alpha.1-12-gaf7a1a9
      CommitHash: af7a1a9
      BuildDate:  2021-11-03T11:38:43Z

== Why a subcommand and not a `-v` or `-V` flag?

Because it can be confusing. Some applications are using `-v` for
version, some are using `-V` for version because `-v` is the verbose
flag. Go, kubectl, and Helm follows the same logic, they have a
`version` subcommand.

**Different flags**

* I find it annoying to parse version information from `version`
  subcommands all the time when I want to automate something, that's the
  `--short`.
* I find it useful to print out everything we have for debugging or
  filing bug reports, that's the `--long` flag.
* I find it useful to print out the package/application name, so even if
  the binary was renamed I can see what is the name of the application,
  but without extra information like commit hash or build date.  That's
  the no-flag option.

**Moved the log.Info call**

In general, it's useful to print out the info log on `run` and `gw`, but
not for `version`.

Fixes #163 

**Checklist**:

### DIFF
--- a/internal/command/gw/gw.go
+++ b/internal/command/gw/gw.go
@@ -15,6 +15,7 @@ import (
 	mvmv1 "github.com/weaveworks/flintlock/api/services/microvm/v1alpha1"
 	cmdflags "github.com/weaveworks/flintlock/internal/command/flags"
 	"github.com/weaveworks/flintlock/internal/config"
+	"github.com/weaveworks/flintlock/internal/version"
 	"github.com/weaveworks/flintlock/pkg/flags"
 	"github.com/weaveworks/flintlock/pkg/log"
 )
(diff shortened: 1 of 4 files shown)


jmickey:
I prefer the version subcommand rather than a flag, and it remains consistent with other CLIs in our ecosystem, like `kubectl`. So 👍 from me.

jmickey (approved):

PR #218, opened 2021-11-05
Refactor the cli functions to use urfave/cli instead of cobra.

### LINKED ISSUE #81: Change to urfave/cli
Change the CLI infra to use [urfave/cli](https://github.com/urfave/cli) instead of cobra/viper. 

This is to make the future deployment option of being combined with containerd easier.

souleb:
Signed-off-by: Soule BA <soule@weave.works>

kind/refactor

Refactor the cli functions to use urfave/cli instead of cobra.
Viper is replaced by urfave/cli atlsrc.

A few variable have been renamed to match their default conterparts, e.g.
cfg.CtrKernelSnapshotter to match defaults.ContainerdKernelSnapshotter

this fixes #81 

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests


internal/command/gw/gw.go
```diff
@@ -9,42 +9,55 @@ import (
...
+			}
+
+			err = altsrc.ApplyInputSourceValues(context, inputSource, cmd.Flags)
+			if err != nil {
+				return fmt.Errorf("unable to apply input source with context: %w", err)
+			}
+		}
+
+		logger := log.GetLogger(context.Context)
+		logger.Infof(
```
richardcase, line 51: We could probably move the version output to the root command?

richardcase: Actually how does this work? Is there the equivalent of the persistent pre-run?

internal/command/root.go
```diff
@@ -2,110 +2,143 @@ package command
...
-	} else {
-		viper.AddConfigPath("$HOME/.config/flintlockd/")
+		return nil
 	}
 
-	_ = viper.ReadInConfig()
-}
+	app.Before = func(context *cli.Context) error {
+		// Load the configuration file
+		configPath := context.String("config")
```
richardcase, line 83: I think we should still be using `XDG_CONFIG_HOME` as the base for the location (like we did previously)/

internal/command/run/run.go
```diff
@@ -21,55 +22,76 @@ import (
...
-	}
+			err = altsrc.ApplyInputSourceValues(context, inputSource, cmd.Flags)
+			if err != nil {
+				return fmt.Errorf("unable to apply input source with context: %w", err)
+			}
+		}
 
-	if err := cmdflags.AddHiddenFlagsToCommand(cmd, cfg); err != nil {
-		return nil, fmt.Errorf("adding hidden flags to run command: %w", err)
+		if cfg.ParentIface == "" {
```
richardcase, line 75: Does `urfave/cli` handle required fields and raising an error if its missing?

internal/command/gw/gw.go
souleb: There is no `persistent pre-run` or `persistent flag`. `App.Before` does what `pre-run` does. And we have access to parent flags through the context.

Now I think moving this to root command (app) is a good idea. But it will be shown only when calling that command

internal/command/run/run.go
souleb: yes there is `required: true`, but the way it behave is that you have to set the flag manually, I think the check happens before the `command.before` where we set the flag with the config file, causing it to fail

richardcase:
I need to properly review this

internal/command/run/run.go
Callisto13: is it possible to set a config file and flags at the same time? which takes precedence?

cmd/flintlockd/main.go
```diff
@@ -1,18 +1,17 @@
...
-	rootCmd, err := command.NewRootCommand()
-	if err != nil {
-		log.Fatalln(err)
-	}
+	app := command.NewApp(os.Stdout)
 
-	if err := rootCmd.Execute(); err != nil {
-		log.Fatalln(err)
+	if err := app.Run(os.Args); err != nil {
+		fmt.Fprintf(os.Stderr, "flintlockd: %s\n", err)
```
jmickey, line 14: Should this be `log.Error`

internal/command/gw/gw.go
jmickey: `App.before` on the App root will run on any subcommand right?

```
	// An action to execute before any subcommands are run, but after the context is ready
	// If a non-nil error is returned, no subcommands are run
	Before BeforeFunc
```
https://github.com/urfave/cli/blob/v1.22.5/app.go#L28

Or is that what you're saying? Maybe I'm not understanding

internal/command/run/run.go
souleb: The precedence for flag value sources is as follows (highest to lowest):

1. Command line flag value from user
2. Environment variable (if specified)
3. Configuration file (if specified)
4. Default defined on the flag

internal/command/gw/gw.go
souleb: I'm sorry I forgot the context of this comment. But yeah `App.before` always run.  A subCommand can also declare `command.Before`.

yitsushi:
Is this still in progress ?

richardcase:
Lets close this for now

Issue #213, opened 2021-11-04
Failure to create + reconcile mVM


Callisto13:
**What happened:**
Saw this appear in CI during an e2e run.

I have not poked yet, but I recall seeing a similar `broken pipe` error when I tried to have more than 1 thing try to write to the firecracker socket. If I am correct, then our reconciliation checks need to be more careful

Logs excerpt:

```
time="2021-11-04T12:58:02Z" level=info msg="checking state of microvm" controller=microvm service=firecracker_microvm vmid=ns0/mvm0
time="2021-11-04T12:58:02Z" level=error msg="failed executing plan" controller=microvm execution_id=01FKNGE50XN3136TVXD5T82F7S execution_time="158.484µs" num_steps=0 plan_name=microvm_create_update
time="2021-11-04T12:58:02Z" level=error msg="failed to reconcile vmid ns0/mvm0: executing plan: executing plan steps: creating plan for microvm_create_update: error occurred generating plan: adding microvm create step: checking if step microvm_create should be included in plan: checking if microvm is running: getting instance info: describing firecracker instance: [GET /][503] describeInstance default  &{FaultMessage:}" controller=microvm
time="2021-11-04T12:58:02Z" level=info msg="Starting reconciliation" action=reconcile controller=microvm vmid=ns0/mvm0
time="2021-11-04T12:58:02Z" level=info msg="Generate plan" action=reconcile controller=microvm stage=plan vmid=ns0/mvm0
time="2021-11-04T12:58:02Z" level=info msg="started executing plan" controller=microvm execution_id=01FKNGE50X0HZGT8HWHKNZRZZX plan_name=microvm_create_update
time="2021-11-04T12:58:02Z" level=info msg="checking state of microvm" controller=microvm service=firecracker_microvm vmid=ns0/mvm0
time="2021-11-04T12:58:02Z" level=error msg="failed executing plan" controller=microvm execution_id=01FKNGE50X0HZGT8HWHKNZRZZX execution_time="146.92µs" num_steps=0 plan_name=microvm_create_update
[...]

PR #221, opened 2021-11-05
fix: ensure create plan is executed once

Callisto13:
The plan list must be cleared from memory between runs so that the plan
completes and the state is saved. This keeps us from falling into an
infinite loop.

Done as a function to emphasise importance!

Fixes https://github.com/weaveworks/flintlock/issues/213

You are all very special

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/core/plans/microvm_delete.go
+++ b/core/plans/microvm_delete.go
@@ -56,7 +56,7 @@ func (p *microvmDeletePlan) Create(ctx context.Context) ([]planner.Procedure, er
 		return []planner.Procedure{}, nil
 	}
 
-	p.steps = []planner.Procedure{}
+	p.clearPlanList()
 
 	// MicroVM provider delete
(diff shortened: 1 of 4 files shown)


yitsushi (dismissed):
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

core/plans/microvm_delete.go
```diff
@@ -98,6 +98,14 @@ func (p *microvmDeletePlan) Result() interface{} {
 	return nil
 }
 
+// This is the most important function in the codebase DO NOT REMOVE
+// Without this, the Create will always return the full origin list of steps
```
yitsushi, line 102: nit: the Delete will always return

Callisto13: copy-pasta fail

yitsushi (approved):

PR #223, opened 2021-11-08
refactor: move thinpool setup to general test Setup/Teardown

Callisto13:
I had this being done in bash via a Docker Entrypoint + `trap`
combination, but it was quite brittle, and sometimes the teardown would
not run which then meant you had to run a manual step to remove thinpools
and loop devices which was annoying.

This changes brings that into the test so the setup and teardown can be
handled programatically.

There is still a case where some manual cleanup may be required: when the tests
fail and devices cannot be removed. I have a thing on my list to have an aggressive
cleanup of all mvms etc on failure. Coming soon!

There is also a case where some "child" pool devices are not cleaned up. I have deliberately
not done anything about this as we should actually fix the code that is leaving those lying
around after a failed delete.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests


richardcase (dismissed):

Callisto13:
ffs sorry @richardcase can i get a re-click?

test/e2e/utils/runner.go
```diff
@@ -73,9 +87,78 @@ func (r *Runner) Teardown() {
 		r.containerdSession.Terminate().Wait()
 	}
 
+	cleanupThinPools()
+	cleanupDirectories()
+
 	gexec.CleanupBuildArtifacts()
 }
 
+func makeDirectories() {
```
yitsushi, line 96: I wonder if we can add this to a `pkg` or somewhere, and reuse in `cmd/dev-helper` and the runner, we could manage this setup from one place.

Callisto13: gonna delete `cmd/dev-helper` instead 😈

yitsushi: It would be still better to manage those settings in one place, I like the Go version and we can use this as a replacement for the shell script (and update the documentation to use that instead)

Callisto13: oh like you mean instead of all the copy-pastable commands we have in the quick-start docs?

for sure, would be good to have a "setup all my stuff" command for people.

can i do that later tho :p

yitsushi (approved):

PR #224, opened 2021-11-08
Add action to run e2es on PRs

Callisto13:
The e2es when run in docker are currently light and fast enough to be
run as part of a PR cycle, so it would be useful to do that.

Not merging for a sec as I look into something weird.


Callisto13:
Closing as is not possible to run firecracker on github action machines: https://github.com/actions/virtual-environments/issues/183

Issue #90, opened 2021-09-15
Resource cleanup ADR


richardcase:
We need to decide how we handle unused images, network interfaces.

When a microvm is deleted the snapshots will be deleted. This may leave a container image unused. What should we do with these unused images? A couple of options:

1. Do nothing
2. Delete them after X days
3. Delete them as soon as they become unused.

richardcase:
I like the idea of providing the command as well and leaving it up to the user to schedule its execution.

LutzLange:
How about an option in the capi object that selects how cleanup as handled? This could end up in the miroVM details and thus mark how quickly cleanup should happen.

yitsushi:
I think the issue with the cleanup in the microvm detail, we need cleanup from leftovers, things we couldn't/didn't delete after vm was deleted, like a snapshot. At that point, we don't have the vmspec anymore only leftover garbage on the system. So it's more like a "garbage collector" rather than a "clean up".

(or i misunderstood the microvm details part)

yitsushi:
Still valid.

PR #226, opened 2021-11-08
feat: Use random generated device name for ifaces

### LINKED ISSUE #171: Reconcile error: `failed to lookup network interface <foobar>_tap: numerical result out of range`
**What happened:**
Saw that the mVM I had previously deleted had survived in some form and that on restart `flintlockd` was trying to reconcile it, but failing with the following error:

```bash
root@fcba54414a45:/src/flintlock# ./bin/flintlockd run --containerd-socket=/run/containerd-dev/containerd.sock --parent-iface=eth0
INFO[0000] flintlockd, version=undefined, built_on=undefined, commit=undefined
INFO[0000] flintlockd grpc api server starting
INFO[0000] starting microvm controller
INFO[0000] starting microvm controller with 1 workers    controller=microvm
INFO[0000] resyncing microvm specs                       controller=microvm
INFO[0000] Resyncing specs                               action=resync controller=microvm namespace=ns
INFO[0000] Starting reconciliation                       action=resync controller=microvm namespace=ns vmid=test_ns_0/test_mvm_0
INFO[0000] Generate plan                                 action=resync controller=microvm namespace=ns stage=plan vmid=test_ns_0/test_mvm_0
INFO[0000] started executing plan                        controller=microvm execution_id=01FJXY6JQJG5EK80ZMH3SSS6JC plan_name=microvm_create
INFO[0000] checking state of microvm                     controller=microvm service=firecracker_microvm vmid=test_ns_0/test_mvm_0
INFO[0000] checking state of microvm                     controller=microvm service=firecracker_microvm vmid=test_ns_0/test_mvm_0
ERRO[0000] failed executing plan                         controller=microvm execution_id=01FJXY6JQJG5EK80ZMH3SSS6JC execution_time=9.026933ms num_steps=4 plan_name=microvm_create
ERRO[0000] failed to resync microvms: resync reconcile for spec test_ns_0/test_mvm_0: executing plan: executing plan steps: executing steps: executing step network_iface_create: checking if networking interface exists: getting interface test_ns_0_test_mvm_0_tap: failed to lookup network interface test_ns_0_test_mvm_0_tap: numerical result out of range  controller=microvm
FATA[0000] starting microvm controller: resyncing specs on start: resyncing microvms: resync reconcile for spec test_ns_0/test_mvm_0: executing plan: executing plan steps: executing steps: executing step network_iface_create: checking if networking interface exists: getting interface test_ns_0_test_mvm_0_tap: failed to lookup network interface test_ns_0_test_mvm_0_tap: numerical result out of range
```

- The mVM to have been fully deleted
- `flintlockd` should not be trying to reconcile it and should be able to start successfully

**How to reproduce it:**
This only happens sometimes, I am not sure yet what the magic thing is which makes this standard procedure go weird.
- Start `containerd` and `flintlockd`.
- Create MicroVM with [example config](https://github.com/weaveworks/flintlock/blob/main/hack/scripts/payload/CreateMicroVM.json)
- Delete MicroVM... or so I thought
- Stop `flintlockd` and `containerd`
- Restart `containerd` and `flintlockd`
- Observe the reconcile error and `flintlockd` stopping

`/var/lib/flintlock` is empty, so nothing to see there.

```bash
root@fcba54414a45:/src/flintlock# ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
198: eth0@if199: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default
    link/ether 02:42:ac:19:00:02 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

```bash
root@fcba54414a45:/src/flintlock# dmsetup ls
dev-thinpool-c  (253:0)
dev-thinpool-c-snap-2   (253:2)                        # <-- root device is still there
```

```bash
root@fcba54414a45:/src/flintlock# ctr --address=/run/containerd-dev/containerd.sock --namespace=flintlock snapshots ls
KEY PARENT KIND
# the snapshot... isn't which is weird, because containerd is logging about it:

DEBU[2021-10-26T09:23:54.667335772Z] get snapshot mounts                           key=flintlock/test_ns_0/test_mvm_0/root
DEBU[2021-10-26T09:23:54.667360793Z] mounts                                        key=flintlock/3/flintlock/test_ns_0/test_mvm_0/root
DEBU[2021-10-26T09:23:54.668455885Z] stat snapshot                                 key="sha256:10bbe4023c58ac8257b9d37b7af1f7293c003be8267cd8416b1a2c24af4757cc"
DEBU[2021-10-26T09:23:54.669290369Z] get snapshot mounts                           key=flintlock/test_ns_0/test_mvm_0/kernel
DEBU[2021-10-26T09:23:54.671413179Z] stat snapshot                                 key="sha256:10bbe4023c58ac8257b9d37b7af1f7293c003be8267cd8416b1a2c24af4757cc"
DEBU[2021-10-26T09:23:54.672587348Z] get snapshot mounts                           key=flintlock/test_ns_0/test_mvm_0/initrd
```

```bash
root@fcba54414a45:/src/flintlock# ctr --address=/run/containerd-dev/containerd.sock --namespace=flintlock leases ls
ID                             CREATED AT           LABELS
flintlock/test_ns_0/test_mvm_0 2021-10-26T08:57:51Z -
```
I will add more details as I find them.

**Environment:**

- flintlock version:

yitsushi:

To fix #171 

Heavily inspired by [Docker -> netutils/utils.go][1]

[1]: https://github.com/docker/docker-ce/blob/1093a93b336461032352e776893afefc2cf3a50d/components/engine/libnetwork/netutils/utils.go#L126

Fixes #171 

**Checklist**:

- [x] includes documentation
- [x] adds unit tests


yitsushi:
In the ADR as a consequence I suggest to raise the priority of #90 to `important-soon`.

core/steps/network/interface_create.go
```diff
@@ -68,10 +69,19 @@ func (s *createInterface) Do(ctx context.Context) ([]planner.Procedure, error) {
 	}
 
 	if s.status == nil {
-		s.status = &models.NetworkInterfaceStatus{}
```
yitsushi, line 71: That's not the way how pointers work.

docs/adr/0005-network-device-name.md
```diff
@@ -0,0 +1,49 @@
...
+value. Docker-ce tries to generate a name 3 times, if the generated name is
+already taken. To reduce possible failures, we decided to retry 5 times,
+it's still not slow, but gives more opportunities on machines with a lot of
+network devices.
+
+[^3]: https://github.com/docker/docker-ce/blob/1093a93b336461032352e776893afefc2cf3a50d/components/engine/libnetwork/netutils/utils.go#L126
+
+## Consequences
+
+It is not possible to determine which MicroVM is the owner of the network
```
Callisto13, line 39: i am sure you have already considered this, but is there no way to save the device name on the mvm state?

yitsushi: We save it in the Status field of the Spec, but from the device name (`/dev/something` or `ip link show`) alone we don't know which vm.

yitsushi: That's not entirely true, but I would say it's a huge hack and can be used for debugging or in the cleanup:
```
❯ cat /proc/$(sudo lsof /dev/tap$(cat /sys/class/net/flvtap6396803/ifindex) 2>/dev/null | awk '/firecrack/ { print $2 }')/cmdline
firecracker--api-sock/var/lib/flintlock/vm/ns1/mvm1/firecracker.sock--idns1-mvm1--boot-timer
```

Callisto13: > And it can be used only if the firecracker instance is running.

ah yeh, true

infrastructure/network/utils.go
```diff
@@ -0,0 +1,64 @@
...
+
+	switch ifaceType {
+	case models.IfaceTypeTap:
+		devPrefix = fmt.Sprintf("%s%s", prefix, tapPrefix)
+	case models.IfaceTypeMacvtap:
+		devPrefix = fmt.Sprintf("%s%s", prefix, macvtapPrefix)
+	case models.IfaceTypeUnsupported:
+		return "", interfaceErrorf("unsupported interface type: %s", ifaceType)
+	default:
+		return "", interfaceErrorf("unsupported interface type: %s", ifaceType)
```
Callisto13, line 44: why have 2 returns for the same case here?

infrastructure/network/utils.go
```diff
@@ -0,0 +1,64 @@
+package network
```
Callisto13, line 1: are there some tests for these bits?

infrastructure/network/utils.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func generateRandomName(prefix string) (string, error) {
+	id := make([]byte, randomBytesLength)
+	if _, err := io.ReadFull(rand.Reader, id); err != nil {
+		return "", err
+	}
+
+	return prefix + hex.EncodeToString(id)[:ifaceLength], nil
+}
```
Callisto13, line 31: teeny nit: can we move unexported funcs to the end?

infrastructure/network/utils.go
yitsushi: The first one for explicit value, the second is a safe-guard if new type is added, catch that too.

yitsushi: updated the second one now say "unknown" not "unsupported"

infrastructure/network/utils.go
yitsushi: done

core/plans/microvm_delete.go
```diff
@@ -122,8 +122,11 @@ func (p *microvmDeletePlan) addStep(ctx context.Context, step planner.Procedure)
 func (p *microvmDeletePlan) addNetworkSteps(ctx context.Context, vm *models.MicroVM, networkSvc ports.NetworkService) error {
 	for i := range vm.Spec.NetworkInterfaces {
 		iface := vm.Spec.NetworkInterfaces[i]
+		ifaceStats := vm.Status.NetworkInterfaces[iface.GuestDeviceName]
```
richardcase, line 125: It shouldn't happen but is it worth checking that the status was found with `ok`?

core/steps/network/interface_create.go
richardcase: And we create the status if its not found:

https://github.com/weaveworks/flintlock/blob/main/core/plans/microvm_create_update.go#L143

docs/adr/0005-network-device-name.md
richardcase: We could also add a utility command (a bit like the cleanup command we are going to add) to dump the vmspec, network interface name etc

core/plans/microvm_delete.go
richardcase: Actually, reading further down i see the test in the `ShouldDo`, so ignore the above

core/steps/network/interface_create.go
richardcase: The code must've been hung over from a previous life.

richardcase (approved):
Looks goot to me. Only 1 comment about adding a utility command for operators

infrastructure/network/utils.go
```diff
@@ -0,0 +1,64 @@
...
+		return "", interfaceErrorf("unsupported interface type: %s", ifaceType)
+	default:
+		return "", interfaceErrorf("unknown interface type: %s", ifaceType)
+	}
+
+	for i := 0; i < retryGenerate; i++ {
+		name, err := generateRandomName(devPrefix)
+		if err != nil {
+			continue
+		}
```
Callisto13, line 42: i don't think the retry generation bit (or the failure of it) is covered by the tests, although i grant you it would be a tricky one to test 😆  so maybe that's fine

Callisto13 (dismissed):

infrastructure/network/utils.go
yitsushi: I rather not create a whole struct and mock it only to test when netlink throws an error. Same goes to the "what happened when io.ReadFull(rand.Reader, id) throws an error".

(I know it's not covered, my nvim tells me :wink: )

Callisto13: 😱 what plugin do you use for that?!? Here i am reading the code like a sucker 🤦‍♀️

yitsushi:
> We could also add a utility command (a bit like the cleanup command we are going to add) to dump the vmspec, network interface name etc

@richardcase : I think it's similar to the one I suggested for the devmapper to place the whole devmapper setup into pkg, that way the e2e test and dev-helper can use it from the same source, we can remove the shell script from the quick-start guide and replace it with a `fldev-helper` or `flsetup` call.

I don't think it is scope either the code changes or the ADR to making a decision how and where do we want to put it if we want to make a utility tool.

Callisto13 (approved):

infrastructure/network/utils.go
yitsushi: `vim-go` has a `:GoCoverageToggle` command.

PR #231, opened 2021-11-09
fix: set mac address for macvtap if specified

richardcase:

There is a problem with using an auto generated mac address with a macvtap interface. The
kernel appears to change the mac address from what it was originally
created with. The firecracker and cloud init configuration contain the original mac address.
When the guest boots and cloud-init gets to configure the network the
mac address has changed and this then causes the network configuration to fail as it can't
match on the mac address.

We need to find actual evidence of the above, but this has been a known
behaviour with other types of interface and for the those the advice is
to specify a mac address when you create the interface.

This change will use the `guest_mac` if it's specified for the mac
address on the host if the interface type is macvtap. For macvtap the
mac address on the host is the same as in the guest.

Also removed an erroneous comment about a block of code not being hit.

Fixes #

**Checklist**:

### DIFF
--- a/core/steps/network/interface_create.go
+++ b/core/steps/network/interface_create.go
@@ -88,10 +88,6 @@ func (s *createInterface) Do(ctx context.Context) ([]planner.Procedure, error) {
 		return nil, fmt.Errorf("checking if networking interface exists: %w", err)
 	}
 	if exists {
-		// This whole block is unreachable right now, because
-		// the Do function is called only if ShouldDo returns
-		// true. It retruns false if IfaceExists returns true.
-		// Line 76 will never return exists=true
 		details, err := s.svc.IfaceDetails(ctx, deviceName)
 		if err != nil {
 			return nil, fmt.Errorf("getting interface details: %w", err)
--- a/infrastructure/network/network_service.go
+++ b/infrastructure/network/network_service.go
@@ -3,6 +3,7 @@ package network
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


infrastructure/network/network_service.go
```diff
@@ -80,14 +90,15 @@ func (n *networkService) IfaceCreate(ctx context.Context, input ports.IfaceCreat
 		return nil, fmt.Errorf("creating interface %s using netlink: %w", link.Attrs().Name, err)
 	}
 
-	macIf, err := netlink.LinkByName(link.Attrs().Name)
+	macIf, err := netlink.LinkByName(input.DeviceName)
 	if err != nil {
 		return nil, fmt.Errorf("getting interface %s using netlink: %w", link.Attrs().Name, err)
```
Callisto13, line 95: is this error still correct? `input.DeviceName` now?

Callisto13 (approved):
zero tests for this file 😆  so adding an issue to backfill

Callisto13 (dismissed):

infrastructure/network/network_service.go
richardcase: I changed it back as there was no need to change that part. And with the original its more obvious that we are re-getting the link we just added.

jmickey (dismissed):

core/steps/network/interface_create.go
```diff
@@ -88,10 +88,6 @@ func (s *createInterface) Do(ctx context.Context) ([]planner.Procedure, error) {
 		return nil, fmt.Errorf("checking if networking interface exists: %w", err)
 	}
 	if exists {
-		// This whole block is unreachable right now, because
-		// the Do function is called only if ShouldDo returns
-		// true. It retruns false if IfaceExists returns true.
-		// Line 76 will never return exists=true
```
yitsushi, line 94: It's still a dead code here and I left this comment it here:
1. It's visible we try to do that and in case we call Do without ShouldDo, it works
2. If the device status is not update and we someone checks the code "but hey we do an update here", they can see "nope we don't" and don't start debugging for hours to get to the same results that it's not executed if it's called only on ShouldDo==true

Callisto13 (approved):

PR #238, opened 2021-11-09
chore(fmt): Enable disabled linters in global scope

### LINKED ISSUE #168: Enable all the disabled linters in global scope
Now we have a lot of errors if we enable `wsl` for example. I put them on ignore/disabled state.

Linters are there to make the Go ecosystem more consistent.

The list of disabled linters:

* `lll`: Reports long lines
* `gci`: Gci control golang package import order and make it always deterministic.
* `godox`: Tool for detection of FIXME, TODO and other comment keywords
* `gochecknoglobals`: check that no global variables exist
* `wsl`: Whitespace Linter - Forces you to use empty lines!

The important part is `wsl`. We should avoid globals (`gochecknoglobals`) and very long lines (`lll`). `gci` would be nice, while `godox` is questionable.

* :exclamation:  Both `wsl` and `lll` makes the code **much** easier to read. 

For `dogox`, I prefer to use issues to track Todo or Fixme items. If it's in the code it will not be fixed until someone has to touch that code block again, while without this linter it's easier to merge a pr that has a comment like
* `TODO: do we need to add validation?` add validation, file an issue, start a discussion.
* `TODO: remove this fatal`, just go and remove that Fatal. If it's not easy, file an issue.
* `TODO: validate the spec`, if it's out of the scope for the issue, but still want to capture "that's the place" than it's enough to write `Here we can validate the spec. #12345`, potentially with an already files issue id

:question:  Maybe with a "what linter settings are we using" ADR.

follow-up from: #162

yitsushi:

Global options:
* An explanation for all nolint annotations is required. Please do not just ignore it, tell us why is it ignored.

**wsl**

* Fixed everywhere.

**lll**

* Fix everywhere.
* Added exception on `https://` links.

**gochecknoglobals**

* Fixed everywhere.

**gci**

* Fixed everywhere (with `golangci-lint run --fix`)

**godox**

No `TODO` comments should live in the code without filed issues to track
them. The reason is simple: if we have a comment with "todo", it has the
same value as not having that comment at all, because no one will care
about it.

* No `TODO`, `BUG`, and `FIXME` comments in the code.
* Except if it has a filed GitHub reference.

**Disabled linters:** [see commit message](https://github.com/weaveworks/flintlock/pull/238/commits/ff5e33ad092e88d64f49249e3192c6f993e7d47a)

Fixes #168 

Removed some comments, because they are not relevant in the code yet. For example add CSI/CNI/RateLimit, we don't have a ticket about it, it's on a long term list, and when we add them, we can extend those structs (and that will basically part of the ticket), right now those comments had no value.

* ADR: work in progress.

**Checklist**:

- [ ] includes documentation


.golangci.yml
```diff
@@ -45,73 +64,68 @@ issues:
...
+  - source: "// .* #\\d+"
+    linters:
+    - godox
+  - path: test/e2e/
+    linters:
+    - goerr113
+    - gomnd
 
 linters:
   disable-all: true
```
richardcase, line 76: Should we flip this and set `enable-all: true` so we pick up new linters in the future? And then document any we explicitly disable

richardcase (approved):
Just the 1 comment

.golangci.yml
yitsushi: `enable-all: true` is deprecated

yitsushi: enabled all and disabled a selection of linters.

richardcase (dismissed):

richardcase (approved):

PR #239, opened 2021-11-09
chore(test): Add configurable opts to e2e tests

Callisto13:
I have added the initial frame for custom test flags.
So far just the options to:
- Skip setting up the devmapper thinpool
- Skip deleting VMs
- Skip all teardown steps
- Set log level for containerd
- Set log level for flintlockd

Unfortunately custom test flags can only be added on `init()` which is
bleh but I don't see another nice was of doing this.

These flags can only be used while running `go test ...` directly.
This means you cannot pass in flags as part of a `make` or a
`docker run -it ...` command, but you can just call `go test etc` from inside the
container or in the metal host and configure what you like.

My next step is some more work around the python tooling so that these
(and more) options can be passed up.

part of https://github.com/weaveworks/flintlock/issues/146

### DIFF
--- a/.github/workflows/nightly_e2e.yml
+++ b/.github/workflows/nightly_e2e.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'pull_request' && github.repository_owner == 'weaveworks' }}
     env:
       PROJECT_NAME: "flintlock_nightly_e2e"
     name: e2e tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -332,42 +332,7 @@ make test
 
 ### Running the end to end tests
 
-There are several ways to run the end to end tests.
-
-#### In your local environment
-
-```
-make test-e2e
-```
-
-This will run the tests directly on your host with minimal fuss.
-You must ensure that you have installed all the dependencies per the
-[Quick-Start guide][quick-start].
-
-#### In a local docker container
-
-```
-make test-e2e-docker
-```
-
-This will run the tests in a Docker container running on your host machine.
-Note that due to the nature of flintlock, the container will be run with
-high privileges and will share some devices and process memory with the host.
-
-#### In an Equinix device
-
-```bash
-export METAL_AUTH_TOKEN=<your token>
-export EQUINIX_ORG_ID=<your org id>
-make test-e2e-metal
-```
-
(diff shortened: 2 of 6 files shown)


Callisto13:
@yitsushi is the test failure i am seeing ^^ the thing that a commit in your iface pr is going to fix?

yitsushi:
> @yitsushi is the test failure i am seeing vv the thing that a commit in your iface pr is going to fix?

Hopefully. I did not see errors yet on my branch. And it has much better error message instead of a huge timeout panic.

yitsushi (approved):

.github/workflows/nightly_e2e.yml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     runs-on: ubuntu-latest
+    if: ${{ secrets.METAL_AUTH_TOKEN != '' || secrets.EQUINIX_ORG_ID != '' }}
```
Callisto13, line 11: derp this should be an `&&`

i hve docs to add so will edit later

Callisto13, line 11: derp this should be an `&&`

i hve docs to add so will edit later

Callisto13, line 11: derp this should be an `&&`

i hve docs to add so will edit later

Callisto13, line 11: derp this should be an `&&`

i hve docs to add so will edit later

Callisto13:
ffs since i cannot comment inline apparently: the action update should have been an `&&` not an `||` but i will fix later when i add docs

.github/workflows/nightly_e2e.yml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     runs-on: ubuntu-latest
+    if: ${{ secrets.METAL_AUTH_TOKEN != '' && secrets.EQUINIX_ORG_ID != '' }}
```
yitsushi, line 11: Tried to add a similar check on the docs publish job, but secrets can't be used in condition.

https://github.com/actions/runner/issues/520

yitsushi: This logic is a valid workaround:

```
    if: ${{ github.event_name != 'pull_request' && github.repository_owner == 'weaveworks' }}
```

Callisto13: boooo stack overflow lied to me 😭 

okay ty

yitsushi (approved):

Issue #198, opened 2021-11-03
VM Supervisor


jmickey:
The Supervisor will be responsible for monitoring running MicroVMs and react to changes that drift from the desired state.

**Why do we need this?**

We don't currently continuously monitor the state of running VMs. If a VM drifts away from its desired state - e.g. A VM crashes and is in a failed state - we need to wait until the next time the reconciler runs for the VM to be recreated/restarted. 

Additionally, we don't currently track if a VM is continuously failing. Flintlock will continue to recreate the VM every time a resync occurs. The reconciler doesn't know if a VM has already been started, as far as it is concerned it only cares about reconciling the existing state to the desired state.

**What do we need**

- The VM Supervisor should exist as a background process/goroutine. 
- As firecracker does not have a background daemon or event system, the supervisor will need to have knowledge of the VMs that should/do exist and their desired state, and should continuously (on a short timer) check the status of each VM.
- The supervisor should probably utilise the containerd state to store events (e.g. VM stopped, VM started, VM restarted X times, etc), and the event bus to notify the reconciler to take action outside the delayed reconciler resync loop.

How this looks on an implementation level is unknown, and it's likely one or more ADRs will need to be produced as a result.

**Subtasks**

- [ ] Create supervisor that starts when `flintlockd` is started, detect and continuously list all VMs and their state directly from Firecracker.
- [ ] Extend MicroVM model to track microvm events - started, stopped, reboots, etc.
- [ ] Detect when VMs are in a failed state and trigger an event on the event bus to reconcile the VM state
- [ ] Track number of failures/restarts and emit metrics.

richardcase:
I don't think the supervisor itself will store events......just raise them. And  it probably shouldn't make any modifications to the microvm spec....so its a read-only consumer of the specs. wdyt?

jmickey:
@richardcase The usage of `Events` here might be a little overloaded. By "Events" I am more referring to a running kind of log I guess? Similar to the events that are shown when you `kubectl describe` a resource. An "event" in this case might be that the supervisor has detected that a previously running VM is no longer running, if that makes sense?

Maybe that is actually part of the reconciler, wdyt?

I don't think the supervisor should make changes to the spec, but maybe the status? Again, maybe I didn't think it through enough and it actually belongs in the reconciler.

jmickey:
[...]

PR #255, opened 2021-11-12
feat: Add retry counter on failed reconciliation

### LINKED ISSUE #178: Flintlock should know when a mVM dies
**What happened:**
An unrelated config issue caused the firecracker process to die shortly after boot. It appeared that flintlock did not notice this (at least according to the logs).

Flintlock should notice that an mVM had died, and it should at least log that this has happened.

**How to reproduce it:**
Going off [this commit](https://github.com/weaveworks/flintlock/commit/28138da238b361bc423abe905db9663b8d00e0ae) (nothing special about it, just head of `main` rn and happens to contain an issue. you could also repro with a new commit and manually kill the pid probably) create a mVM with the config in `hack/scripts/payload/CreateMicroVM.json`.

You should see the firecracker process start, then after a few moments die. The firecracker logs will talk about no init found, but that is not important here. The issue is that flintlock did not appear to notice.

 We do not have Get merged right now, so curious to know what the status knows.

**Environment:**

- flintlock version:

### LINKED ISSUE #232: Add retry counter on failed reconciliation
Right we we have code to abort reconciliation for a VM if it failed N times, but we never set the Retried field.

If something is wrong, we get an infinite reconciliation loop. It would be nice if the system would stop doing it after N failures and clean the Retries field on successful reconciliation.

nothing

yitsushi:

If reconciliation failed, increase the Retry counter, set the NotBefore
field to a future date and reschedule a retry.

A Go routine handles the force retry because the system tries to
reconcile only if an event tell the system to do that or with the fixed
periodical Resync (which is slow for that).

Because we never tracked if a MicroVM was able to boot or not, we just
let the reconciler to check if the process is not there and react to the
results. In case the MicroVM was not able to boot, we reported back a
success on the MicroVM start step, which is not right and we can't track
failed state with that. As a solution, now a step has a Verify function
that will be called after Do. If the result is false, it marks the step
failed. That way we can start the MicroVM, wait a bit and check if it's
still running, if it's not running, the start failed.

Fixes #232 
Fixes #178

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


yitsushi:
Did not add test on `reconsile.go`, we have zero tests for it and I think it's out of scope to write the full test for it, and filed an issue #264

core/application/reconcile.go
```diff
@@ -52,52 +57,99 @@ func (a *app) ResyncMicroVMs(ctx context.Context, namespace string) error {
...
+		return nil
 	}
 
+	plan := a.plan(spec, localLogger)
+
 	execCtx := portsctx.WithPorts(ctx, a.ports)
 
 	executionID, err := a.ports.IdentifierService.GenerateRandom()
 	if err != nil {
+		if scheduleErr := a.reschedule(ctx, localLogger, spec); scheduleErr != nil {
```
richardcase, line 138: If we can't get a random number should we re-schedule? Seems like that wouldn't be a transient error we should fail straight away.

core/steps/microvm/start.go
```diff
@@ -63,3 +70,24 @@ func (s *startStep) Do(ctx context.Context) ([]planner.Procedure, error) {
...
 	return nil, nil
 }
+
+func (s *startStep) Verify(ctx context.Context) error {
+	logger := log.GetLogger(ctx).WithFields(logrus.Fields{
+		"step": s.Name(),
+		"vmid": s.vm.ID,
+	})
+	logger.Debug("waiting for the microvm to start")
+	time.Sleep(time.Duration(s.bootTime) * time.Second)
```
richardcase, line 80: I don't think we should wait here. We can check the state as long as its running (and maybe configured) then we are good. If it fails later in the boot process, then the supervisor or periodic resync can find this out.

richardcase: I do like the idea of `Verify` though :+1:

yitsushi: We don't have supervisor yet and the periodic resync is slow. It fails like in 3s, the next resync happens minutes later.

core/application/reconcile.go
yitsushi: I don't understand this. What is the random number?

yitsushi: Ok I see :rofl:

yitsushi: I'm not sure, even if it seems a transient error, if it's a problem on the host, all vms will reconcile in an infinite loop. Reschedule prevent this with increasing the Retry value by one. If it's a transient error, the next execution will happen in 20s, but we don't get an infinite loop in case something is wrong with the host.

core/steps/microvm/start.go
yitsushi: I agree this check should live in the supervisor, but #198 is not there yet.

core/application/reconcile.go
richardcase: My point is that i don't see this being a transient error in this situation, and so retrying up to the maximum number of retries won't change anything. Can we fail fast and stop reconciliation?

core/steps/microvm/start.go
richardcase: I think it's still good to have a check here, but sleeping for 30 seconds is too much. If i send a request in to create 5 microvms this will add 2.5 minutes to the provisioning time.

I think the `sleep` should be completely removed. And if needed we can bump up the priority of the supervisor.

If we change the sleep time to say 5s then it may fail after 6s.

yitsushi: > sleeping for 30 seconds is too much
oops I planned to do 5s not 30s.

I see. If it dies on boot, it should die in 2-3 seconds. If it dies later it may died for different reasons and most likely it's not a Start problem. I don't see how can we detect boot error without waiting a bit.

core/application/reconcile.go
yitsushi: So like terminate flintlock if this happens?

core/steps/microvm/start.go
yitsushi: Removing the sleep, maybe we need a new Restart counter? If we called `start.Do()` when the knows Status from the spec is not Running, increase that counter. I don't know.

core/application/reconcile.go
richardcase: It feels like it shouldn't continue to run as i don't see a situation where it would start working again. Can you see this ever being a transient error which will be corrected on retry?

And maybe we try generating a random number early using the service before we start the reconciler? wdyt?

core/steps/microvm/start.go
richardcase: It's not ideal, whatever we do here. If we sleep for 5 seconds the boot may fail after this still. Without the supervisor or an active health check from an agent running in the guest there are always going to be situations where we have to wait for the resync to detect a failure.

The Start action API call can fail as well.....i've had this quite a few times.

Yes i think  we do need to distinguish between failing to reconcile and restarting because a microvm has failed.....including when reporting metrics.

yitsushi: My point on the 5s sleep is not to detect if the VM has issues in long term, but to prevent infinite loop when the VM can't boot. If it fails after 5s that's fine, but if it can't live for at least 5s it's around 99% the next 10 minutes we will get 100000 lines of error messages because it always tried to reconcile and the system always said "yes I did it".

core/application/reconcile.go
yitsushi: Well, reading ulid, I think we can get error if the time is too big or the entropy increment hits overflow. I don't think any of these can be resolved without restarting the application or even the server (not the resync). The rest is when it was not able to read from bytes that can happen is the system is very busy and can't read from urandom because of some kind of io error, that can be resolved over time.

core/steps/microvm/start.go
richardcase: Ok, so sleep for 5 seconds and then come back an review/remove the sleep when we have the supervisor?

yitsushi: Sounds good.

core/application/reconcile.go
richardcase: Thats great to understand the scenarios where it may fail, thanks for looking at this....great :+1:

core/models/microvm.go
```diff
@@ -59,6 +59,8 @@ type MicroVMStatus struct {
 	NetworkInterfaces NetworkInterfaceStatuses `json:"network_interfaces"`
 	// Retry is a counter about how many times we retried to reconcile.
 	Retry int `json:"retry"`
+	// DeletedAt indicates the time the microvm was marked as deleted.
```
richardcase, line 62: nit: comment is out of sync

core/steps/microvm/start.go
```diff
@@ -13,16 +14,22 @@ import (
 	"github.com/weaveworks/flintlock/pkg/planner"
 )
 
-func NewStartStep(vm *models.MicroVM, vmSvc ports.MicroVMService) planner.Procedure {
+func NewStartStep(
+	vm *models.MicroVM,
+	vmSvc ports.MicroVMService,
+	bootTime int,
```
richardcase, line 20: nit: maybe `bootWaitTimeSeconds` or something more descriptive.....just idea and feel free to ignore.

pkg/planner/planner.go
```diff
@@ -25,4 +25,6 @@ type Procedure interface {
 	Do(ctx context.Context) ([]Procedure, error)
 	// ShouldDo determines if this procedure should be executed
 	ShouldDo(ctx context.Context) (bool, error)
+	// Verify the state after Do.
```
richardcase, line 28: nit: might be worth adding some more context as to why we might want to verify and that its only for certain situations.

Again, a nice to have but not essential.

richardcase (commented):
Just a few nit...feel free to resolve them depending on what you think.

core/steps/microvm/start_test.go
```diff
@@ -44,25 +44,32 @@ func TestNewStartStep(t *testing.T) {
 	ctx := context.Background()
 	vm := testVMToStart()
 
-	step := microvm.NewStartStep(vm, microVMService)
+	step := microvm.NewStartStep(vm, microVMService, 1)
```
Callisto13, line 47: can we var these `1`s here? just so it is easy to know at a glance what this param is meant to be

core/models/microvm.go
```diff
@@ -59,6 +59,8 @@ type MicroVMStatus struct {
 	NetworkInterfaces NetworkInterfaceStatuses `json:"network_interfaces"`
 	// Retry is a counter about how many times we retried to reconcile.
 	Retry int `json:"retry"`
+	// NotBefore tells the system to do not reconsile until given timestamp.
```
Callisto13, line 62: teeny nit: `reconcile` with a `c`

core/steps/microvm/start_test.go
```diff
@@ -164,3 +170,42 @@ func TestNewStartStep_ServiceStartError(t *testing.T) {
...
+		EXPECT().
+		State(ctx, vm.ID.String()).
+		Return(ports.MicroVMStateUnknown, nil)
+
+	subSteps, err := step.Do(ctx)
+	verifyErr := step.Verify(ctx)
+
+	g.Expect(subSteps).To(g.BeEmpty())
+	g.Expect(err).To(g.BeNil())
+	g.Expect(verifyErr).To(g.MatchError(internalerr.ErrUnableToBoot))
```
Callisto13, line 200: can we group these a little clearer?

```suggestion
	subSteps, err := step.Do(ctx)
	g.Expect(err).To(g.BeNil())
	g.Expect(subSteps).To(g.BeEmpty())
	
	verifyErr := step.Verify(ctx)
	g.Expect(verifyErr).To(g.MatchError(internalerr.ErrUnableToBoot))
```

core/application/reconcile.go
```diff
@@ -52,52 +57,99 @@ func (a *app) ResyncMicroVMs(ctx context.Context, namespace string) error {
...
+	return plans.MicroVMCreateOrUpdatePlan(input)
+}
+
+func (a *app) reschedule(ctx context.Context, logger *logrus.Entry, spec *models.MicroVM) error {
+	spec.Status.Retry++
+	waitTime := time.Duration(spec.Status.Retry*backoffBaseInSeconds) * time.Second
+	spec.Status.NotBefore = time.Now().Add(waitTime).Unix()
+
+	logger.Infof(
+		"[%d/%d] reconciliation failed, reschedule at %s",
```
Callisto13, line 88:
```suggestion
		"[%d/%d] reconciliation failed, rescheduled for next attempt at %s",
```
bit clearer?

core/application/reconcile.go
```diff
@@ -52,52 +57,99 @@ func (a *app) ResyncMicroVMs(ctx context.Context, namespace string) error {
...
+	)
+
+	if _, err := a.ports.Repo.Save(ctx, spec); err != nil {
+		return fmt.Errorf("saving spec after plan failed: %w", err)
+	}
+
+	go func(id, ns string, sleepTime time.Duration) {
+		logger.Info("Wait to emit update")
+		time.Sleep(sleepTime)
+		logger.Info("Emit update")
```
Callisto13, line 101: do we want to be a bit more clear on these logs?

like "waiting to publish update event" or "waiting to reschedule for update"

core/application/reconcile.go
```diff
@@ -52,52 +57,99 @@ func (a *app) ResyncMicroVMs(ctx context.Context, namespace string) error {
...
+
+		_ = a.ports.EventService.Publish(
```
Callisto13, line 103: do we not care about logging this err? could imagine that being an annoying one to solve if it failed

core/application/reconcile.go
```diff
@@ -52,52 +57,99 @@ func (a *app) ResyncMicroVMs(ctx context.Context, namespace string) error {
...
 	}
 
+	plan := a.plan(spec, localLogger)
+
 	execCtx := portsctx.WithPorts(ctx, a.ports)
 
 	executionID, err := a.ports.IdentifierService.GenerateRandom()
 	if err != nil {
+		if scheduleErr := a.reschedule(ctx, localLogger, spec); scheduleErr != nil {
+			return fmt.Errorf("saving spec after plan failed: %w", scheduleErr)
```
Callisto13, line 139: more relevant error message? (ditto line 150)

core/application/reconcile.go
yitsushi: I don't think we even need these lines, added them while i was working on it.

core/application/reconcile.go
yitsushi: Yes, intentionally ignored it. This go routine is a steroid on the system, if it fails next resync will do the trick. But we can log. Hopefully ppl don't get confused reading the log because it happens "random" as it's not in the flow (async go routine).

Callisto13 (approved):
nice work 🎉

PR #256, opened 2021-11-12
Fix tabbing in release workflow

Callisto13:
This was my bad, i am having a hell of a day with indentation 😄 , but the tabbing there was a bit off so i formatted the whole file.


yitsushi (approved):
It's funny how the `on` section was recognized as "ah it's only indentation" and the rest "hell no, they are completely different lines"

PR #263, opened 2021-11-15
chore(test): Tidy up python tool

Callisto13:

Only the second commit does anything new, the rest is tidying up.

The main purpose of this PR is to fix the e2e nightly build. I think I must
have messed up a rebase or something because I had run the exact
command so many times and it was fine so 🤷‍♀️ .

I also did a bit of tidying up in the `test/tools/` dir. i have no idea how
a python project should be structured (even a toy one like this) but it was a
bit messy in there and there is much room for refactoring and tests so
organising things better now felt sensible.


test/tools/config/config.py
```diff
@@ -8,7 +8,7 @@
 class Config:
     def __init__(self):
         self.dir = dirname(abspath(__file__))
-        self.base = dirname(dirname(dirname(abspath(__file__))))
+        self.base = dirname(dirname(dirname(dirname(abspath(__file__)))))
```
yitsushi, line 11: Maybe with `pathlib.Path` it's more readable:
```
from pathlib import Path
```

```suggestion
        self.base = str(Path(abspath(__file__)).parent.parent.parent.parent)
```

yitsushi (approved):
Small style suggestion, not a blocker. lgtm

test/tools/config/config.py
yitsushi: In theory, `parents[4]` should work too, but then we have to check the length first

Callisto13: > Maybe with `pathlib.Path` it's more readable

maybe, i will add that with my next pr in all the places we do this 👍

PR #265, opened 2021-11-16
feat: Add script to setup direct-lvm mode for containerd

Callisto13:

Using loop devices for devmapper's thinpool is not recommended for
production.

Real block devices are required to back thinpools in production.

This PR:
- adds an alternative script which users can run if they want a
  production-ready setup. The script will take a given free device and
  create the required physical and logical volumes, formatting them into
  thinpool storage.
- this device can also be provided via an environment variable, which
makes it easier to use the script in the tooling userdata
- updates the quickstart docs to include the option.
- adds the script to the equinix tooling userdata so that equinix vms
  are bootstrapped with direct-lvm by default.
- changes the default machine created by the tooling so that a type
  with a spare disk which is not mounted or partitioned is available. This
  saves us having to sort out extra storage which is a right pain now that
  Equinix has deprecated their native storage offering.

Part of https://github.com/weaveworks/flintlock/issues/146

There is more wiring with the tests coming soon, just separating out the PRs

**Checklist**:

- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/hack/scripts/direct_lvm.sh
+++ b/hack/scripts/direct_lvm.sh
@@ -1,0 +1,139 @@
+#!/bin/bash
+
+# WARNING: THIS SCRIPT HAS MUTLIPLE PURPOSES.
+# TAKE CARE WHEN EDITING.
+
+# This is a scripted version of https://docs.docker.com/storage/storagedriver/device-mapper-driver/#configure-direct-lvm-mode-manually
+
+set -ex
+
+if [[ $(id -u) != 0 ]]; then
+  echo "Run this script as root..." >&2
+  exit 1
+fi
+
+DISK_NAME="$THINPOOL_DISK_NAME"
+POOL_NAME_PREFIX="flintlock"
+
+usage() {
+cat << EOF
+usage: $0 -d DISK_NAME [-p POOL_NAME]
+
+Script to generate release data for eksctl and profiles projects.
+
+OPTIONS:
+   -d YEAR	  (Required if \$THINPOOL_DISK_NAME not set) Set the name of the disk to use for thinpool storage
+   -p POOL_NAME	  Name of the thinpool to create (default: flintlock)
+   -h		  Show this message
+EOF
+}
+
+while getopts ":hd:p:" OPTION
+do
+  case $OPTION in
+    h)
+      usage
+      exit
+      ;;
+    d)
+      DISK_NAME=$OPTARG
+      ;;
+    p)
+      POOL_NAME_PREFIX=$OPTARG
+      ;;
+    ?)
+      usage
+      exit
+      ;;
+  esac
+done
+
+if [[ -z "$DISK_NAME" ]]; then
+  echo "one of \$THINPOOL_DISK_NAME or '-d NAME' must be set"
+  exit 1
+fi
+
+DISK_PATH="/dev/$DISK_NAME"
+PROFILE="/etc/lvm/profile/$POOL_NAME_PREFIX-thinpool.profile"
+CROOT=/var/lib/containerd-flintlock
+DIR="${CROOT}/snapshotter/devmapper"
+
+echo "will create thinpool $POOL_NAME_PREFIX-thinpool on $DISK_PATH"
+
+apt update
+apt install -y thin-provisioning-tools lvm2
+
+if [[ $(pvdisplay) != *"$DISK_PATH"* ]]; then
+  pvcreate "$DISK_PATH"
+  echo "created physical volume on $DISK_PATH"
+fi
+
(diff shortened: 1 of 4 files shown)


hack/scripts/direct_lvm.sh
```diff
@@ -0,0 +1,96 @@
...
+# This is a scripted version of https://docs.docker.com/storage/storagedriver/device-mapper-driver/#configure-direct-lvm-mode-manually
+
+set -ex
+
+if [[ $(id -u) != 0 ]]; then
+  echo "Run this script as root..." >&2
+  exit 1
+fi
+
+DISK_NAME="${1:-sdb}"
```
yitsushi, line 15: I think if disk name was not defined, we should exit, that can lead to very serious issues if called without an argument, I'm not sure I would be happy if someone executed this script on my server where `sdb` is NOT empty.

yitsushi (changes requested):

hack/scripts/direct_lvm.sh
Callisto13: yeh good point.

i did it like this to be easy with the bootstrap.sh, as passing a flag there would be annoying 🤔

richardcase:
Very interesting, trying this now.

hack/scripts/direct_lvm.sh
```diff
@@ -0,0 +1,96 @@
...
+DISK_PATH="/dev/$DISK_NAME"
+POOL_NAME="${1:-flintlock}"
+PROFILE="/etc/lvm/profile/$POOL_NAME-thinpool.profile"
+CROOT=/var/lib/containerd-dev
+DIR="${CROOT}/snapshotter/devmapper"
+
+apt update
+apt install -y thin-provisioning-tools lvm2
+
```
richardcase, line 24: Be good to have some `echo` logging giving feedback as it runs.

hack/scripts/direct_lvm.sh
```diff
@@ -0,0 +1,96 @@
...
+set -ex
+
+if [[ $(id -u) != 0 ]]; then
+  echo "Run this script as root..." >&2
+  exit 1
+fi
+
+DISK_NAME="${1:-sdb}"
+DISK_PATH="/dev/$DISK_NAME"
+POOL_NAME="${1:-flintlock}"
```
richardcase, line 17: This is using the same arg (i.e. 1) as the disk name, so the volume group ends up being called the same as the disk and not flintlock.

richardcase: Guessing it should be `POOL_NAME="${2:-flintlock}"`

hack/scripts/direct_lvm.sh
```diff
@@ -0,0 +1,96 @@
...
+PROFILE="/etc/lvm/profile/$POOL_NAME-thinpool.profile"
+CROOT=/var/lib/containerd-dev
```
richardcase, line 19: Should the `/var/lib/containerd-dev` be different when using this? Perhaps `/var/lib/containerd-flintlock`?

hack/scripts/direct_lvm.sh
Callisto13: ah yes excellent catch thank you!

hack/scripts/direct_lvm.sh
Callisto13: yep sounds good

hack/scripts/direct_lvm.sh
Callisto13: i have `-x` set so it is pretty loud, but yeh i can echo some things too

richardcase:
I'm getting snapshot failures when i use a lvm pool with flintlock when running locally (with a spare empty disk): `failed to query device metadata for "flintlock-thinpool-snap-5": not found`.....investigating.

Callisto13:
huh interesting, haven't had an issue on equinix after around 6 runs. don't have a spare disk to validate locally i am afraid. add commits to this pr if it's not you 👍

Callisto13:
There is also a link at the top of the script to where I got the commands from, so check there incase they have any troubleshooting etc

richardcase:
> huh interesting, haven't had an issue on equinix after around 6 runs of the e2es. don't have a spare disk to validate locally i am afraid. add commits to this pr if it's not you +1

Pretty certain the issues are entirely of my own making

hack/scripts/direct_lvm.sh
richardcase: Looks like this good after removing the default disk name

Callisto13: 👍  got distracted doing other things for a while 😄 

pr comment changes coming up soon

yitsushi (approved):

PR #269, opened 2021-11-17
Review OCI tests

### LINKED ISSUE #15: Review OCI tests

yitsushi:

* use yitsushi/devmapper-containerd-action@v1
* extends imageservice test
  * invalid image pull
  * double check if volume is mounted
  * volume is not mounted test
  * kernel volume mount test
* full unit test on ImageService

fixes #15

* That's the maximum I can do with integration tests.
* There is one branch I couldn't test with unit test, it would require some magical refactor because I can't fake this private method to return true as it has an internal variable that's updated inside the the same private method.
```
func snapshotExists(ctx context.Context, key string, ss snapshots.Snapshotter) (bool, error) {
	snapshotExists := false

	err := ss.Walk(ctx, func(walkCtx context.Context, info snapshots.Info) error {
		if info.Name == key {
			snapshotExists = true
		}

		return nil
	})
	if err != nil {
		return false, fmt.Errorf("walking snapshots: %w", err)
	}

	return snapshotExists, nil
}
```

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests


infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,215 @@
...
+func getTestVolumeImage() string {
+	envImage := os.Getenv("CTR_TEST_VOL_IMG")
+	if envImage != "" {
+		return envImage
+	}
+
+	return testImageVolume
+}
+
+func getTestKernelImage() string {
```
Callisto13, line 190: this func isn't being used, I'm guessing you wanted it in place of all the spots where you have `testImageKernel`?

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,215 @@
...
+		testOwnerNamespace,
+		testOwnerName,
+		testOwnerUsageID,
+	)
+	expectedLeaseName := fmt.Sprintf("flintlock/%s/%s", testOwnerNamespace, testOwnerName)
+
+	defer func() {
+		// Make sure it's deleted.
+		client.ImageService().Delete(namespaceCtx, testImageKernel)
+		client.ImageService().Delete(namespaceCtx, testImageVolume)
```
Callisto13, line 66: the volume image used on line 46 might be different than the one being cleaned up, did you mean to use `getTestVolumeImage()` here (and on a few other lines which have `testImageVolume`)?

Callisto13 (commented):
I'm struggling to run these tests a bit. I'm guessing we need to make sure `devpool.sh` has run, and that containerd is running, and that the `CTR_SOCK_PATH` is set. It would be good to document this, or have this setup and teardown automated (the teardown especially can be annoying to remember).

Also when I have that all set up and run `make test-with-cov`, the tests are still skipped, I think because the tests are executed with `sudo` in the make target and the env is not preserved (bear in mind that we cannot do `sudo make test-with-cov` in the action because then they will be run with a different Go version 🤦‍♀️ )

Callisto13:
Looking at the action logs these integration tests are not running:

```
    event_service_test.go:25: skipping containerd event service integration test
--- SKIP: TestEventService_Integration (0.00s)
=== RUN   TestImageService_Integration
    image_service_integration_test.go:31: skipping containerd image service integration test
```

Callisto13:
Can we add some comments or log lines to each stage of the test, just so that it is easier to tell at a glance what should be happenning (will be useful later when/if something breaks to know exactly what should be going on and if the test is legit doing that). Something like:
```go
// When <condition X> <thing Y> should happen
```

yitsushi:
Log lines sounds a good idea, I'll add them.

Missing test run, interesting, maybe I don't set an environment variable with my action.

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,215 @@
...
+		client.SnapshotService(testSnapshotter).Remove(namespaceCtx, expectedSnapshotName)
+		leases, _ := client.LeasesService().List(namespaceCtx)
+		for _, lease := range leases {
+			client.LeasesService().Delete(namespaceCtx, lease)
+		}
+	}()
+
+	err := imageSvc.Pull(ctx, inputGet)
+	Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
yitsushi: I like "called code" and "test what we got" lines separated from each other.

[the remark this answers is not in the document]
Callisto13: 👍

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,291 @@
...
+// == Chapter III: Kernel
+//
+// We do the same checks for kernel. The kernel and volume works the same way,
+// and the real reason why we test it here because the usage type is different,
+// and we should be able to pull and mount them.
+//
+// == Chapter IV: Delete
+//
+// We should be able to delete both the volume and the kernel image. After we
+// delete them, we should see zero images in containerd.
```
Callisto13, line 75: can we re-word this to explain we are testing?

from what i can see, the deletion is not us so we don't really care much about it. the test from what i can see is actually on `Exists`

honestly the whole delete bit could be moved to cleanup as it doesn't have anything to do with us and we already tested Exist earlier.

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,291 @@
...
+//
+// == Chapter II: Snapshots
+//
+// As we already pull and mounted an image, we expect one snapshot to be there
+// with the name constructed from the VM namespace, the VM name, and the volume
+// usage type.
+// At the same time, we expect one lease to be there with similar naming
+// structure except the volume usage type we have one lease for all resources
+// in containerd.
+//
```
Callisto13, line 65: can we clarify that this is not a separate test stage in itself. we are not checking any new code here, it is part of the PullAndMount checks

Callisto13: actually it could be moved after the kernel Pull and then we can check that all the snapshots are what we expect like in a loop or smth?

infrastructure/containerd/image_service_integration_test.go
Callisto13: actually, where does our code call various Deletes on the containerd services? I expected to see it around here but can't 🤔

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,291 @@
...
+	expectedLeaseName := fmt.Sprintf("flintlock/%s/%s", testOwnerNamespace, testOwnerName)
+
+	defer func() {
+		// Make sure it's deleted.
+		client.ImageService().Delete(namespaceCtx, getTestKernelImage())
+		client.ImageService().Delete(namespaceCtx, getTestVolumeImage())
+		client.SnapshotService(testSnapshotter).Remove(namespaceCtx, expectedSnapshotName)
+		leases, _ := client.LeasesService().List(namespaceCtx)
+		for _, lease := range leases {
+			client.LeasesService().Delete(namespaceCtx, lease)
```
Callisto13, line 121: i am not sure this is working. i have a lease left over after running

when i wrap with 
```suggestion
			Expect(client.LeasesService().Delete(namespaceCtx, lease)).To(Succeed())
```
it does not error so 🤷‍♀️ 

(actually might be good to do Succeed check like that just so we don't get silent failures)

infrastructure/containerd/image_service_test.go
```diff
@@ -2,133 +2,669 @@ package containerd_test
...
+	err := client.Pull(ctx, &ports.ImageSpec{
+		ImageName: testImage,
+		Owner:     testOwner,
+	})
+	g.Expect(err).ToNot(g.HaveOccurred())
+}
+
+// TestImageService_Pull_failedLease tests what happens when something goes
+// wrong with Leases.
+func TestImageService_Pull_failedLease(t *testing.T) {
```
Callisto13, line 59: missing test manager.Create failure in getExistingOrCreateLease

infrastructure/containerd/image_service_test.go
```diff
@@ -2,133 +2,669 @@ package containerd_test
...
-	client, ctx := testCreateClient(t)
-	namespaceCtx := namespaces.WithNamespace(ctx, testContainerdNs)
+	err := client.Pull(ctx, &ports.ImageSpec{
+		ImageName: testImage,
+		Owner:     testOwner,
+	})
+	g.Expect(err).To(g.HaveOccurred())
+}
+
+// TestImageService_Pull_failedLease tests a successful PullAndMount.
```
Callisto13, line 88:
```suggestion
// TestImageService_PullAndMount tests a successful PullAndMount.
```

infrastructure/containerd/image_service_test.go
```diff
@@ -2,133 +2,669 @@ package containerd_test
...
+		LeasesService().
+		Return(leasesManager)
 
-	client, ctx := testCreateClient(t)
-	namespaceCtx := namespaces.WithNamespace(ctx, testContainerdNs)
+	err := client.Pull(ctx, &ports.ImageSpec{
+		ImageName: testImage,
+		Owner:     testOwner,
+	})
+	g.Expect(err).To(g.HaveOccurred())
```
Callisto13, line 85: let's check that the error is the one we expect here (ditto all the rest too probably)

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+type testImageMountOptions struct {
+	ImageName string
+	Owner     string
+	Use       models.ImageUse
+	Expected  bool
+}
+
+// testImageMounted checks if we can mount an image.
```
Callisto13, line 281:
```suggestion
// testImageMounted checks that images have been mounted
// tests the IsMounted method
```

Callisto13:
@yitsushi do you know why we call `withOwnerLease` on line 65 of `PullAndMount` and then pass the context into `imageExists` and then maybe `pullImage` which then both call `withOwnerLease` themselves again? Seems a bit redundant?

infrastructure/containerd/image_service_integration_test.go
yitsushi: https://github.com/weaveworks/flintlock/pull/269/files/113cb403c180621e966325abae9d7776bf09e65a#diff-d96f2d206636039c31d85ae360ad7739ae7487bd2c5847daffa4db30446d0135R215-R219

yitsushi: So it's:
1. delete images
2. check if they are there (one by one)
3. check List, it should be zero.
```
	//
	// Chapter IV: Delete
	//

	err = client.ImageService().Delete(namespaceCtx, getTestKernelImage())
	Expect(err).NotTo(HaveOccurred())

	err = client.ImageService().Delete(namespaceCtx, getTestVolumeImage())
	Expect(err).NotTo(HaveOccurred())

	exists, err = imageSvc.Exists(ctx, &ports.ImageSpec{
		ImageName: getTestKernelImage(),
		Owner:     testOwnerUsageID,
	})
	Expect(err).ToNot(HaveOccurred())
	Expect(exists).To(BeFalse())

	images, err := client.ImageService().List(namespaceCtx)
	Expect(err).NotTo(HaveOccurred())
	Expect(len(images)).To(BeZero())
```

infrastructure/containerd/image_service_integration_test.go
yitsushi: In the defer cleanup section I would never put test. That's there to ignore everything just try to delete. We have delete test at the end of the test function. This defer call is there in case the test was interrupted/cancelled/killed/whatevercanhappen.

infrastructure/containerd/image_service_integration_test.go
Callisto13: `client.ImageService().Delete` is not our code right? we are just calling containerd directly?

infrastructure/containerd/image_service_integration_test.go
Callisto13: it doesn;t need to be a "test" test, just so that if things are left around because deletes fail we actually know about it, rather than being surprised when the next test run hits a conflict

yitsushi:
> do you know why we call withOwnerLease on line 65 of PullAndMount and then pass the context into imageExists and then maybe pullImage which then both call withOwnerLease themselves again? Seems a bit redundant?

(pure assumption)
For those separate function, it's because `imageExists` and `pullImage` are called from other places like `Pull` for `pullImage` or `IsMounted` for `imageExists`.
Calling `withOwnerLease` in `PullAndMount`, I don't know maybe to make sure we can get the Lease and prevent it to take unnecessary steps. I have no idea.

infrastructure/containerd/image_service_test.go
yitsushi: I wouldn't go that deep as they can be very ugly by design as we just pass through the error without checking anything on them and at the end is a wrapped error wrapped into a new error that's wrapped into a new error that's wrapped ................. and who knows how deep is that. From the point we call `client.Pull` the error is basically the same containerd will throw back and we just wrap them with `fmt.Errorf(..., err)`.

yitsushi: That would be cool if we can refactor this code to be less chaotic (and a bit more complex) on error handling, but with the code we have now, I don't think it has any use to test if the error is `nope` wrapped in N layers.

Callisto13: you can do like a `ContainSubstring` and look for a certain piece

Callisto13: but yeh LOTS of refactoring would be good in all this area

infrastructure/containerd/image_service_integration_test.go
yitsushi: The logic is there to mimic the usage. (we actually don't even use `Pull` only `PullAndMount`, it's just an extra test). So `Chapter I` is a full workflow we are using on CreateVM on one volume. The Kernel does the same, but it does some bits different, but overall the same, so that test was added to make sure the PullAndMouont part is working on kernel images too, but the rest handled the same.

infrastructure/containerd/image_service_integration_test.go
Callisto13: where do we actually call delete on things? like in the code not in tests? i can't find where flintlock deleted containerd things on mvm teardown. or wait do we not do that yet? is that one of the things we left for later? yes i am remembering now we have a thing for it. nvm as you were

yitsushi: The I would say just remove that part and keep the defer cleanup section as it is. I see no reason to test if it has any errors or not, not out code and the connection should be fine otherwise the rest of the test would fail anyway.

infrastructure/containerd/image_service_integration_test.go
yitsushi: The issue with `Expect` it stops when it encounters an error. That's why I added the defer block there. if the test fails anywhere, the rest will not be called, and in this case the cleanup does it without any error handling, just do it.

infrastructure/containerd/image_service_test.go
yitsushi: > you can do like a `ContainSubstring` and look for a certain piece

well, it's the same if I saw "occurred" because I know it will contain what I tell it to use
```
		Return(nil, errors.New("nope"))
```
and we have zero custom error handling in this module. Actually we have one in content store, but this is the ImageService.

infrastructure/containerd/image_service_integration_test.go
Callisto13: 👍  sounds good

and yeh i expected that we would have more of a delegation "shim" around containerd (and its various services) rather than just having the client and using it directly. maybe we can move towards that eventually with some refactoring, as it could lead us to a cleaner pattern which would be easier to test in-depth

infrastructure/containerd/image_service_integration_test.go
yitsushi: So the only reasonable addition to this would be
```
if err := client.LeasesService().Delete(namespaceCtx, lease); err != nil {
  // log the error.
}
```
I can add some logging in cleanup.

infrastructure/containerd/image_service_integration_test.go
Callisto13: > we actually don't even use Pull

lol! why are we keeping it? for a rainy day?

infrastructure/containerd/image_service_integration_test.go
Callisto13: aaah i see, yeh good point

infrastructure/containerd/image_service_test.go
Callisto13: don't we wrap it? if we hit the failure at that point i would expect to have substring `"listing existing containerd leases: nope"`

infrastructure/containerd/image_service_integration_test.go
yitsushi: maybe :rofl:

infrastructure/containerd/image_service_test.go
yitsushi: yes we wrap, but we wrap all of them. so it will be always `something: something: something: something: nope` (and the amount of `something` changes based on what we called), but 100% it will contain `nope`.

Callisto13: so we can still do
```suggestion
	g.Expect(err).To(g.MatchError(ContainSubstring("listing existing containerd leases: nope")))
```

yitsushi: (pure personal opinion) In my eyes it falls into the "we can test it for coverage and because looks good, but useless"

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,281 @@
...
+	inputGet.ImageName = getTestKernelImage()
+
+	err = imageSvc.Pull(ctx, inputGet)
+	Expect(err).NotTo(HaveOccurred())
+
+	exists, err := imageSvc.Exists(ctx, &ports.ImageSpec{
+		ImageName: getTestVolumeImage(),
+		Owner:     testOwnerUsageID,
+	})
+	Expect(err).ToNot(HaveOccurred())
```
Callisto13, line 208:
```suggestion
	Expect(err).NotTo(HaveOccurred())
```
:trollface:

infrastructure/containerd/image_service_integration_test.go
```diff
@@ -0,0 +1,281 @@
...
+// testImageMounted checks that images have been mounted
+// tests the IsMounted method
+func testImageMounted(ctx context.Context, imageSvc ports.ImageService, opts testImageMountOptions) {
+	mounted, err := imageSvc.IsMounted(ctx, &ports.ImageMountSpec{
+		ImageName:    opts.ImageName,
+		Owner:        opts.Owner,
+		Use:          opts.Use,
+		OwnerUsageID: testOwnerUsageID,
+	})
+	Expect(err).ToNot(HaveOccurred())
```
Callisto13, line 279:
```suggestion
	Expect(err).NotTo(HaveOccurred())
```
:trollface: :trollface:

infrastructure/containerd/image_service_test.go
Callisto13: i mean it doesn;t look _great_ :trollface: 

but seriously 90% of the reviews i have done in my life have been "do we not want to catch that error?" followed by an "oh shit yeh!" response. so checking errors are returned when we expect and in the way we expect them is an unhealthy obsession of mine 😭

Callisto13 (approved):
LGTM, just some final trolls

infrastructure/containerd/image_service_integration_test.go
yitsushi: But but but but bu but... `to not have occurred` vs `not to have occurred`, the first one sounds better to my ears.

Callisto13: honestly they both sound off to me, i have had this discussion like 100 times 😂  i have a vim shortcut which autocompletes to NotTo and I have no idea why

everywhere else we have `NotTo` so for consistency i guess?

yitsushi:
```
flintlock on  15-review-oci-tests via  v1.17.1
❯ ag 'NotTo' | wc -l
97

flintlock on  15-review-oci-tests via  v1.17.1
❯ ag 'ToNot' | wc -l
31

flintlock on  15-review-oci-tests via  v1.17.1
❯ ag -l 'NotTo'
core/models/vmid_test.go
core/application/app_test.go
core/steps/runtime/dir_create_test.go
core/plans/microvm_delete_test.go
core/plans/microvm_create_update_test.go
test/e2e/utils/utils.go
test/e2e/utils/runner.go
pkg/wait/wait_test.go
pkg/log/log_test.go
pkg/validation/validate_test.go
pkg/planner/actuator_test.go
infrastructure/containerd/repo_test.go
infrastructure/containerd/image_service_integration_test.go
infrastructure/controllers/microvm_controller_test.go
infrastructure/network/utils_test.go

flintlock on  15-review-oci-tests via  v1.17.1
❯ ag -l 'ToNot'
core/models/vmid_test.go
core/steps/runtime/kernel_mount_test.go
core/steps/runtime/initrd_mount_test.go
core/steps/runtime/repo_release_test.go
core/steps/runtime/volume_mount_test.go
core/steps/microvm/start_test.go
core/steps/microvm/create_test.go
core/steps/microvm/delete_test.go
core/steps/network/interface_create_test.go
core/steps/network/interface_delete_test.go
core/steps/event/publish_test.go
test/e2e/utils/utils.go
test/e2e/e2e_test.go
infrastructure/containerd/image_service_test.go
infrastructure/containerd/image_service_integration_test.go
```

it's more like 2/3.

Callisto13: lool and some of those ToNots were mine 🤦‍♀️

Callisto13 (approved):

PR #271, opened 2021-11-18
refactor: apply opinionated api defaults based on accepted adr

### LINKED ISSUE #179: Opinionated API Changes
Currently the API is generic. We should consider being more opinionated towards our use case. For example, should the user have to specify a network interface for metadata requests? It may be better that if we see metadata in the request then we automatically creatre a network interface for MMDS requests

jmickey:

- Automatically add the metadata service network device
- Default CPU and RAM to 2 and 1024 respectively
- Apply the default namespace is non is provided
- Separate Volumes into RootVolume and AdditionalVolumes

Fixes #179

**Checklist**:

- [x] includes documentation
- [x] adds unit tests


core/models/vmid.go
```diff
@@ -31,7 +31,7 @@ func NewVMID(name, namespace string) (*VMID, error) {
 	}
 
 	if namespace == "" {
-		return nil, coreerrs.ErrNamespaceRequired
+		namespace = "default"
```
yitsushi, line 34: Can we put store this in a constant, so other packages can use this as value and we have to update only one place?

core/models/volumes.go
```diff
@@ -6,7 +6,10 @@ type Volume struct {
 	ID string `json:"id"`
 	// IsRoot specifies that the volume is to be used as the root volume. A machine
 	// must have a root volume.
-	IsRoot bool `json:"is_root"`
+	//
+	// NOTE: This field does not exist on the gRPC API, it is automatically filled
+	// when converting from the API model to the this internal model.
+	// IsRoot bool `json:"is_root"`
```
yitsushi, line 12: Is this comment `//` intentional on the Volume model? This is the internal model. Either the comment on the field or the commented out field seems odd.

infrastructure/grpc/convert.go
```diff
@@ -26,6 +26,15 @@ func convertMicroVMToModel(spec *types.MicroVMSpec) (*models.MicroVM, error) {
 			MemoryInMb: int64(spec.MemoryInMb),
 		},
 	}
+
+	if convertedModel.Spec.VCPU == 0 {
+		convertedModel.Spec.VCPU = 2
```
yitsushi, line 31: Same as the namespace, `conts`?

infrastructure/grpc/convert.go
```diff
@@ -26,6 +26,15 @@ func convertMicroVMToModel(spec *types.MicroVMSpec) (*models.MicroVM, error) {
...
+	}
+
+	if convertedModel.Spec.MemoryInMb == 0 {
+		convertedModel.Spec.MemoryInMb = 1024
```
yitsushi, line 35: Same as the namespace and vcpu, `conts`?

yitsushi (commented):
I don't know why `gomnd` did not yell on vcpu and memory default values.

core/models/volumes.go
jmickey: The comment _was_ valid, but I removed the need for `is_root` completely and forget to remove this altogether.

yitsushi (approved):
Small nit: `MemoryInMb` sounds like it's Megabits not Megabytes, but lgtm

PR #272, opened 2021-11-18
feat(test): Check capacity of facility before creating device

Callisto13:
This changes the config so that a list of facilities will be tested to
ensure that there is capacity for device creation.

This gives us some resilience in the nightly e2e tests, although if
there is no capacity anywhere then they will still fail.

I had wanted to use a list of metros instead, which would have left
Equinix to find an appropriate facility, but the package I am using to
hit the Equinix API has a bug in that method (I've put in a PR). I have removed setting the
metro as an option as I did not feel like dealing with that logic right
now. Also I am not really sure which takes precedence if you set both on
the request. Metro I think?

Fixes #

**Checklist**:

part of https://github.com/weaveworks/flintlock/issues/146

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/tools/metal/welder.py
+++ b/test/tools/metal/welder.py
@@ -4,6 +4,7 @@ import spur
 import logging
 import os
 import sys
+from .error import CapacityError
 from Crypto.PublicKey import RSA
 from os.path import dirname, abspath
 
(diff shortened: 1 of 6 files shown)


test/tools/metal/welder.py
```diff
@@ -32,18 +33,30 @@ def set_logger(self, level):
         return logger
 
     def create_all(self):
+        facility = None
+        try:
+            facility = self.check_capacity()
```
yitsushi, line 38: Can we print some debug messages about what facility will be used? It's a useful information and in case Equinix has for example networking issues in one of their facilities we can figure out if the test failure is not related to our changes.

Callisto13: yep good idea 👍

yitsushi (approved):

Callisto13:
okay codeql is hanging and it wont let me cancel and re-run the workflow so i am just gonna merge. did not add any go anyway

PR #295, opened 2021-11-29
Make sure a mvm is properly stopped

### LINKED ISSUE #225: Snapshots sometimes not properly cleaned on delete
**What happened:**
Occasionally on a delete, `containerd` fails to remove a device.

```
msg="failed to remove device" error="failed to deactivate device \"dev-thinpool-snap-2\": failed to deactivate device: device or resource busy"
```

I have seen this in CI and pretty consistently locally.

In CI ([link](https://github.com/weaveworks/flintlock/runs/4123046102?check_suite_focus=true)) the tests eventually failed while checking that the PID was no longer running after delete. (Another instance of that happening [here](https://github.com/weaveworks/flintlock/runs/4184458494?check_suite_focus=true))

Locally the test does not fail sometimes (it does fail more often than not), but the device still remains.

The reconciliation does make some attempt to clean up, but in the case of the local "leak", even though the mVM eventually died, the device stayed.

On a delete we should ensure that everything is gone.

**How to reproduce it:**

- Start containerd
- Start flintlock
- Create mvm: `./hack/scripts/send.sh -m CreateMicroVM`
- Get it `./hack/scripts/send.sh -m GetMicroVM`
- Delete it : `./hack/scripts/send.sh -m DeleteMicroVM`
- Observe the logs from containerd:
   ```
   WARN[2021-11-26T12:23:06.066934932Z] failed to deactivate device, retrying... (10 of 100)  error="device or resource busy"
   WARN[2021-11-26T12:23:55.593179253Z] failed to deactivate device, retrying... (20 of 100)  error="device or resource busy"
   etc ...
   ```
- Get the mvm again: `./hack/scripts/send.sh -m GetMicroVM`
- See that it fails, because according to flintlock the mvm is gone
- Look for the firecracker process: `ps aux | grep fire`
- See that it is not, in fact, gone
- Kill the pid: `sudo kill -9 <pid>`
- See that now the logs continue and containerd is fine, the mvm is actually deleted

**Environment:**

- flintlock version:

souleb:
Signed-off-by: Soule BA <soule@weave.works>

kind/bug

If implemented this will add a wait command to make sure a mvm is
stopped before resuming the deletion plan.

Fixes #225

**Checklist**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/infrastructure/firecracker/provider.go
+++ b/infrastructure/firecracker/provider.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"syscall"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
@@ -25,22 +26,26 @@ type Config struct {
 	StateRoot string
 	// RunDetached indicates that the firecracker processes should be run detached (a.k.a daemon) from the parent process.
 	RunDetached bool
+	// DeleteVMTimeout is the timeout to wait for the microvm to be deleted.
+	DeleteVMTimeout time.Duration
 }
 
 // New creates a new instance of the firecracker microvm provider.
 func New(cfg *Config, networkSvc ports.NetworkService, fs afero.Fs) ports.MicroVMService {
 	return &fcProvider{
-		config:     cfg,
-		networkSvc: networkSvc,
-		fs:         fs,
+		config:          cfg,
+		networkSvc:      networkSvc,
+		fs:              fs,
+		deleteVMTimeout: cfg.DeleteVMTimeout,
 	}
 }
 
 type fcProvider struct {
 	config *Config
 
-	networkSvc ports.NetworkService
-	fs         afero.Fs
+	networkSvc      ports.NetworkService
+	fs              afero.Fs
+	deleteVMTimeout time.Duration
 }
 
 // Capabilities returns a list of the capabilities the Firecracker provider supports.
@@ -105,6 +110,14 @@ func (p *fcProvider) Delete(ctx context.Context, id string) error {
 
 	if sigErr := process.SendSignal(pid, syscall.SIGHUP); sigErr != nil {
 		return fmt.Errorf("failed to terminate with SIGHUP: %w", err)
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(ctx, p.deleteVMTimeout)
+	defer cancel()
+
+	// Make sure the microVM is stopped.
+	if err := process.WaitWithContext(ctxTimeout, pid); err != nil {
+		return fmt.Errorf("failed to wait for pid %d: %w", pid, err)
 	}
 
 	logger.Info("deleted microvm")
--- a/internal/command/run/run.go
+++ b/internal/command/run/run.go
@@ -68,6 +68,7 @@ func NewCommand(cfg *config.Config) (*cobra.Command, error) {
 
 	cmd.Flags().StringVar(&cfg.StateRootDir, "state-dir", defaults.StateRootDir, "The directory to use for the as the root for runtime state.")
 	cmd.Flags().DurationVar(&cfg.ResyncPeriod, "resync-period", defaults.ResyncPeriod, "Reconcile the specs to resynchronise them based on this period.")
+	cmd.Flags().DurationVar(&cfg.DeleteVMTimeout, "deleteMicroVM-timeout", defaults.DeleteVMTimeout, "The timeout for deleting a microvm.")
 
 	return cmd, nil
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,4 +40,6 @@ type Config struct {
 	ResyncPeriod time.Duration
 	// MaximumRetry defined how many times we retry if reconciliation failed.
 	MaximumRetry int
+	// DeleteVMTimeout defines the timeout for the delete vm operation.
+	DeleteVMTimeout time.Duration
 }
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -48,6 +48,9 @@ const (
 	// ResyncPeriod is the default resync period duration.
 	ResyncPeriod time.Duration = 10 * time.Minute
 
+	// DeleteVMTimeout is the default timeout for deleting a microvm.
+	DeleteVMTimeout time.Duration = 10 * time.Second
+
 	// DataDirPerm is the permissions to use for data folders.
 	DataDirPerm = 0o755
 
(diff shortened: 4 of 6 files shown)


internal/config/config.go
```diff
@@ -40,4 +40,6 @@ type Config struct {
 	ResyncPeriod time.Duration
 	// MaximumRetry defined how many times we retry if reconciliation failed.
 	MaximumRetry int
+	// Timeout defines the general timeout for the reconciliation.
+	Timeout time.Duration
```
yitsushi, line 44: Can we add a more descriptive name for that? Reading in the `internal/config` package "Timeout" feels like "timeout for what?"

pkg/defaults/defaults.go
```diff
@@ -48,6 +48,8 @@ const (
 	// ResyncPeriod is the default resync period duration.
 	ResyncPeriod time.Duration = 10 * time.Minute
 
+	Timeout time.Duration = 10 * time.Second
```
yitsushi, line 51: Same, I don't think we are using any "global" timeout value.

yitsushi (commented):
I'm not sure we want to wait for the vm to start, if it did not start, next reconciliation loop (and later supervisor) will catch it. We had a similar discussion about it with @richardcase before.

@richardcase : can you take a look at this PR and share your thoughts?

infrastructure/firecracker/provider.go
```diff
@@ -25,6 +26,8 @@ type Config struct {
 	StateRoot string
 	// RunDetached indicates that the firecracker processes should be run detached (a.k.a daemon) from the parent process.
 	RunDetached bool
+	// Timeout is the timeout for the firecracker process operations.
+	Timeout time.Duration
```
richardcase, line 30: Following on from @yitsushi comment, perhaps this should be a more specific `ShutdownTimeout` or something similar.

internal/command/run/run.go
```diff
@@ -68,6 +68,7 @@ func NewCommand(cfg *config.Config) (*cobra.Command, error) {
 
 	cmd.Flags().StringVar(&cfg.StateRootDir, "state-dir", defaults.StateRootDir, "The directory to use for the as the root for runtime state.")
 	cmd.Flags().DurationVar(&cfg.ResyncPeriod, "resync-period", defaults.ResyncPeriod, "Reconcile the specs to resynchronise them based on this period.")
+	cmd.Flags().DurationVar(&cfg.Timeout, "timeout", defaults.Timeout, "The general timeout for the reconciliation.")
```
richardcase, line 71: Again.....best to make this a specific timeout instead of a generalised one.

pkg/defaults/defaults.go
richardcase: I agree, so this would become the default for the shutdown timeout

richardcase:
> I'm not sure we want to wait for the vm to start, if it did not start, next reconciliation loop (and later supervisor) will catch it. We had a similar discussion about it.

Yes i don't think we need to wait for it to start. We just want to give the process time to gracefully stop, and then kill it if not. We may need timeouts in other scenarios in the future,  so overall i think we should have timeout flags that are specific to a usage (i.e. vm shutdown timeout).

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
+package process
```
Callisto13, line 1: this should be:
```suggestion
package process_test
```

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	//release the pid
+	p.Wait()
+
+	//check if process exists
+	exists, err := Exists(p.Process.Pid)
+	if err != nil {
+		t.Error("error while retrieving process: %w", err)
+	}
```
Callisto13, line 31: gomega matchers here (and everywhere) please:
```suggestion
	g.Expect(err).NotTo(HaveOccurred())
```

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if exists {
+		t.Error("process still exists")
+	}
```
Callisto13, line 35:
```suggestion
	g.Expect(exists).To(BeFalse())
```

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func TestWaitWithContext(t *testing.T) {
+	testCases := []struct {
+		name          string
+		command       []string
+		timeout       int
+		expectedError string
+	}{
+		{
+			name:          "wait on a process that exists",
```
Callisto13, line 46:
```suggestion
			name:          "wait on a process that exists, should succeed",
```

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
...
+			command:       []string{"sleep", "10"},
+			timeout:       5,
+			expectedError: "",
+		},
+		{
+			name:          "wait on a process that does not exist",
```
Callisto13, line 52:
```suggestion
			name:          "wait on a process that does not exist, should error",
```

pkg/process/process_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+			if err != nil {
+				if tc.expectedError != "" {
+					if err.Error() != tc.expectedError {
+						t.Errorf("unexpected error: %s", err)
+					}
+				} else {
+					t.Errorf("unexpected error: %s", err)
+				}
+			}
```
Callisto13, line 83:
```suggestion
			if tc.expectedError != "" {
				g.Expect(err).To(MatchError(tc.expectedError))
			} else {
			        g.Expect(err).NotTo(HaveOccurred())
			}
```

pkg/process/process.go
```diff
@@ -50,7 +51,7 @@ func SendSignal(pid int, sig os.Signal) error {
 		return fmt.Errorf("unable to find process (%d): %w", pid, err)
 	}
 
-	if err := proc.Signal(os.Interrupt); err != nil {
+	if err := proc.Signal(sig); err != nil {
```
Callisto13, line 54: ha! nice catch

Callisto13 (commented):
The code is tidy 🎉 , I just have commented on the tests to line them up with our practices.
I agree with what the others have said.

souleb:
> The code is tidy 🎉 , I just have commented on the tests to line them up with our practices. I agree with what the others have said.

I have slightly changed the test `TestWaitWithContext` because `on Unix system os.FindProcess() is always successful`

Callisto13 (dismissed):

souleb:
@Callisto13 do you mind approving, I just rebased.

richardcase (approved):

PR #304, opened 2021-12-03
chore(deps): Add action to bump version of firecracker in e2es

Callisto13:
Note: The firecracker version included in the bootstrap.sh of this PR
is a dummy version so I can see that everything works.
(Everything code-wise still works, the number is just one behind what
it should be while I set up.)

The action will open a PR to bump the version when a new one is
available upstream.
The new fork firecracker release wont actually exist yet, as there are some
manual steps required to update the macvtap feature branch on our fork.
These manual steps will be detailed in the PR.

See it in action here: https://github.com/weaveworks/flintlock/pull/307


jmickey (approved):

.github/workflows/fork.yaml
```diff
@@ -0,0 +1,77 @@
...
+        if [[ "$new_version" == "" ]]; then
+          echo "No new Firecracker release. Nothing to do. (this is not an error, no worries.)"
+          exit 1
+        fi
+
+        echo 'NEW_VERSION<<EOF' >> $GITHUB_ENV
+        echo "${new_version}" >> $GITHUB_ENV
+        echo 'EOF' >> $GITHUB_ENV
+    - name: Create PR
+      if: ${{ success() }}
```
richardcase, line 29: Instead of `if: ${{ success() }}` i'm wondering if you change the expression so that this is only executed if there is a new version, so something like  `if: ${{ env.NEW_VEWRSIOn != "" }}`  and then in the **Bump firecracker** step we do do away with the `exit 1`

Callisto13: yeh that sounds better 👍

jmickey (approved):

PR #308, opened 2021-12-06
feat: remove option to set volume snapshotter

### LINKED ISSUE #180: Only allow devmapper for volumes
Currently when using a container for a volume we allow the user to specify a different snapshotter instead of devmapper. flintlock won't work easily with another snapshotters for volumes so we should remove this option. 

Callisto13:

Flintlock does not handle other snapshotters well, so we should be
opinionated about this. There is no real reason to support others.

Fixes #180

I have done the lightest possible change here: simply removed the flag and set the default at the point we wire the containerd config. I chose to do this here rather than in the ImageService (there is a little `getSnapshotter` thing) so that all containerd config was grouped upfront for easier tracing.

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


richardcase (dismissed):
I like that its set outside of the image service....just in case another snapshotter comes along that we can use.

richardcase (approved):

PR #320, opened 2021-12-13
feat: Change API to use Additional Kernel Args

### LINKED ISSUE #312: Change API to use Additional Kernel Args

Currently we expect the user to specify the kernel args when creating a new microvm. This requires the user to know what args are recommended for Firecracker (and in the future Cloud Hypervisor).

We should adopt an opinionated view on the kernel cmdline args and use ones that are recommended by FC or cloud Hypervisor and only allow the user to add to them.

We should extend the microvm Provider interface to return the kernel cmdline args.

The user of flintlock shouldn't need to know what kernel args are recommended for a specific provider and so we should hide this to give a better UX.

yitsushi:

 The user of flintlock shouldn't need to know what kernel args are recommended for a specific provider and so we should hide this to give a better UX.

Fixes #312

**Checklist**:

- [x] includes documentation
- [x] adds or updates e2e tests


api/types/microvm.proto
```diff
@@ -66,7 +66,9 @@ message MicroVMSpec {
 message Kernel {
   // Image is the container image to use.
   string image = 1;
-  // Cmdline is the kernel command line args.
+  // Cmdline is the additional kernel command line args. Each provider has its
+  // own recommended list, they will be used automatically. This field is for
+  // additional values.
```
richardcase, line 71: I'm wondering if we change away from a string to a key:value structure (map[string]string in go) where value can be a empty string. And then we construct the cmd line as a string after we we take the defaults for the provider and merge them with the additional ones supplied by the user.

We'd want some sort of precedence logic as well. For example, we default Firecracker to using a data source of no cloud with the Firecracker metadata service but perhaps they have their own metadata service so in the additional args they would supply `ds=nocloud-net;s=http://supercoolmetadata/latest/` and it would overwite the default for firecracker.

infrastructure/firecracker/config.go
```diff
@@ -88,7 +89,11 @@ func WithMicroVM(vm *models.MicroVM) ConfigOption {
 			})
 		}
 
-		kernelArgs := vm.Spec.Kernel.CmdLine
+		kernelArgs := defaults.KernelCmdLine
```
richardcase, line 92: I think the provider should publish their default args (as a k:v)...related to comment above.

infrastructure/firecracker/config.go
```diff
@@ -88,7 +89,11 @@ func WithMicroVM(vm *models.MicroVM) ConfigOption {
...
+
+		if vm.Spec.Kernel.CmdLine != "" {
```
richardcase, line 94: This logic can then be generic (i.e. live outside of the firecracker provider) to flintlock and use the defaults from the provider.

infrastructure/firecracker/config.go
```diff
@@ -113,6 +119,40 @@ func WithMicroVM(vm *models.MicroVM) ConfigOption {
...
+	return config.KernelCmdLine{
+		"console":       "ttyS0",
+		"reboot":        "k",
+		"panic":         "1",
+		"pci":           "off",
+		"i8042.noaux":   "",
+		"i8042.nomux":   "",
+		"i8042.nopnp":   "",
+		"i8042.dumbkbd": "",
+		"ds":            "nocloud-net;s=http://169.254.169.254/latest/",
```
richardcase, line 152: Is it worth adding a comment  about the `ds` arg as well?

richardcase (approved):
I like this, it's good that if CmdLine is specified in the request that it overwrites the values of any of the default args if they exist or just adds a new arg. Adding the ability to remove a default arg would make the api more complex

There is one minor comment but think we can go with this.

infrastructure/firecracker/config.go
yitsushi: I tried to find a kernel level documentation about it, but the only one I found is the cloudinit documentation that does not really tell us what is it really only to "just set it and you are good", and that documentation can be found easily searching for `nocloud-net`.

https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html

PR #321, opened 2021-12-13
chore(test): Get only first output from create func

Callisto13:
Small thing I did not notice at the time.


test/tools/run.py
```diff
@@ -99,7 +99,7 @@ def create_device(config_file, org_id, project_id, device_name, userdata):
         f"Creating device {cfg['device']['name']} with config {pretty}")
 
     welder = Welder(token, cfg)
-    ip = welder.create_all()
+    ip = welder.create_all()[0]
```
richardcase, line 102: Can the call to `welder.create_all()` fail? Although this question has no bearing on the change.

richardcase (approved):
What would

test/tools/run.py
Callisto13: yes it can, but it has no impact on what that value is used for, as i have neglected to handle any errors properly and they just panic 😜

PR #324, opened 2021-12-14
Add service file for systemd

Callisto13:
As part of my docs writing I realised that having an option to manage flintlockd with systemd looks a bit more pro than "start the server with an `&`".

The `--parent-iface` flag is of course required and means some sort of user input. I did consider an EnvironmentFile, but it's a bit "eh" and the systemd creators deeply regret its inclusion.
I am documenting that users need to do something here.

### DIFF
--- a/flintlockd.service
+++ b/flintlockd.service
@@ -1,0 +1,34 @@
+# Copyright The flintlock Authors
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


yitsushi:
We have viper included, i wonder why we don't have a `--config` option.

Callisto13:
are we not removing viper?

Callisto13:
but yeh if urfave/cli has a config file then we can use that

richardcase:
> but yeh if urfave/cli has a config file then we can use that

The PR that Soule has done for urfave/cli does  have a config file option.

Callisto13:
👍  do we want that as part of this PR or can i move forward with this one and come back to that?

richardcase:
> +1 do we want that as part of this PR or can i move forward with this one and come back to that?

Lets come back to it

PR #327, opened 2021-12-17
feat: Use flintlock unique id

### LINKED ISSUE #291: Create globally unique identifier per microvm
Currently, the identifier for a microvm is a combination of namespace and name. We should also create an identifier that is globally/universally unique as well (i.e. a UUID/GUID).  This will stop the unlikely but possible (this has happened in CAPA) scenario where 2 separate management clusters create a cluster with the same name/namespace.

We should also be able to get the microvm details with this id. 

yitsushi:

This commit message will be a QA like commit message

Why in the spec and not somewhere else?

Historically we are using the same spec in create request an responses,
we do not return with status on create request because it's mostly
empty. The uuid could be there, but on create there is no status yet.

Why is it set on create request?

We can't return the uuid in the response if it's not generated and
without the uuid they can't delete the resource, it would be a disaster
if capmvm sends a request to create ns1/mvm1, then it wants to delete it
even before we created the resources. If the delete request would be
only ns1/mvm1 it would delete the old one, or basically any other
microvms that has the same name and namespace (coming from other
management clusters).

Why is it global and not microvm specific?

It seems logical and generating and maintaining a unique id shouldn't be
the job of the microvm provider, their responsibility should be to
manage microvms, metadata and other stuffs are provider independent.

Fixes #291

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


jmickey (dismissed):
LGTM

api/services/microvm/v1alpha1/microvms.proto
```diff
@@ -58,11 +58,13 @@ message CreateMicroVMResponse {
 message DeleteMicroVMRequest {
   string id = 1;
   string namespace = 2;
+  string uuid = 3;
```
richardcase, line 61: I'm not sure about this. I think we need 2 versions of the Delete/Get API methods:
- Delete/Get(name, namespace)
- DeleteById/GetById(id)

I still think its valid to get or delete  a microvm based on its namespace/name

richardcase: This would probably then mean changing `id` to `name` throughout flintlock

core/application/commands.go
```diff
@@ -26,17 +26,25 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
...
-		vmid, err := models.NewVMID(name, defaults.MicroVMNamespace)
+		vmid, err := models.NewVMID(name, defaults.MicroVMNamespace, "")
 		if err != nil {
 			return nil, fmt.Errorf("creating vmid: %w", err)
 		}
 
 		mvm.ID = *vmid
 	}
 
+	uuid, err := a.ports.IdentifierService.GenerateRandom()
```
richardcase, line 37: The infrastructure implementation that we use for the identifier service doesn't generate a UUID but instead a ULID. It does say a ULID is compatible with a UUID but its not formatted as a UUID with this call.

Maybe we need to refer to this as a uid instead of uuid to indicate that its not a UUID? Or something similar.

richardcase (changes requested):

api/services/microvm/v1alpha1/microvms.proto
yitsushi: I wouldn't let users to delete by `ns/name` only because in some cases it will not be only one, same for get. The get endpoint returns one spec, but without the uid, it's not guaranteed that's the only one.

yitsushi: Or we have to update the Get return a list of specs.

core/application/commands.go
yitsushi: actually, it's even more confusing now. Turned out i used `a.ports.IdentifierService.GenerateRandom()` here and in the grpc server `uuid.New().String()`  which one should we keep?

yitsushi: i'll use ulid to me more consistent with the rest of the system.

api/services/microvm/v1alpha1/microvms.proto
yitsushi: I would even go remove everything from the delete and get and leave it only the uid. So you can Get (single spec response) only with uid and delete the same way.

core/application/commands.go
richardcase: Yeah ulid from the identifier service is the best.

api/services/microvm/v1alpha1/microvms.proto
richardcase: Yes i think perhaps you are right. Get and delete should only accept the id.  I think we should also implement (in a different PR) a `Find` that we can use to find by the ns/name.....wdyt?

core/application/commands.go
```diff
@@ -26,17 +26,25 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
...
 		if err != nil {
 			return nil, fmt.Errorf("creating vmid: %w", err)
 		}
 
 		mvm.ID = *vmid
 	}
 
+	uid, err := a.ports.IdentifierService.GenerateRandom()
+	if err != nil {
+		return nil, fmt.Errorf("generating random name for microvm: %w", err)
```
Callisto13, line 39: the nittiest of nits:
```suggestion
		return nil, fmt.Errorf("generating random ID for microvm: %w", err)
```

hack/scripts/payload/DeleteMicroVM.json
```diff
@@ -1,4 +1,5 @@
 {
   "id": "mvm1",
-  "namespace": "ns1"
+  "namespace": "ns1",
+  "uid": "b1c0aa3c-ebba-498a-a7c7-903d91758d83"
```
Callisto13, line 4: should the other 2 be removed?

infrastructure/grpc/server.go
```diff
@@ -46,6 +46,13 @@ func (s *server) CreateMicroVM(
 		return nil, status.Error(codes.InvalidArgument, "invalid request")
 	}
 
+	/*
+		if req.Microvm.Uid == nil {
+			newUID := uuid.New().String()
+			req.Microvm.Uid = &newUID
+		}
+	*/
```
Callisto13, line 54: this meant to be here still?

Callisto13 (approved):

richardcase (approved):

infrastructure/grpc/server.go
yitsushi: nope XD

hack/scripts/payload/DeleteMicroVM.json
yitsushi: yes

yitsushi: But that "tooling" has to be refined after Find.

Callisto13 (approved):

PR #329, opened 2021-12-20
refector: remove pause/result from provider interface

richardcase:

As we are not supporting pausing / resuming microvms at this stage these
functions have been removed from the microvm provider interface. The
firecracker implementation has these removed as well and any tests have
been updated.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/infrastructure/firecracker/provider.go
+++ b/infrastructure/firecracker/provider.go
@@ -2,7 +2,6 @@ package firecracker
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"syscall"
 	"time"
@@ -15,8 +14,6 @@ import (
 	"github.com/weaveworks/flintlock/pkg/log"
 	"github.com/weaveworks/flintlock/pkg/process"
 )
-
-var errNotImplemeted = errors.New("not implemented")
 
 // Config represents the configuration options for the Firecracker infrastructure.
 type Config struct {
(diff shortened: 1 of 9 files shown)


infrastructure/firecracker/provider.go
```diff
@@ -71,19 +68,15 @@ func (p *fcProvider) Start(ctx context.Context, vm *models.MicroVM) error {
...
-}
-
-// Resume will resume a paused microvm.
-func (p *fcProvider) Resume(ctx context.Context, id string) error {
-	return errNotImplemeted
-}
-
-// Stop will stop a paused or running microvm.
+// Stop will stop a running microvm.
 func (p *fcProvider) Stop(ctx context.Context, id string) error {
```
Callisto13, line 72: are we using stop anywhere?

Callisto13: can we not just remove it if it serves no purpose?

richardcase: Why not. I did make me wonder if we should be doing different things in stop and delete. For example, in `Stop` should we stop the firecracker process and then in `Delete` we delete the files. But lets create a ticket for follow up and remove stop for the time being. wdyt?

Callisto13: sounds good 👍

Callisto13 (approved):

infrastructure/firecracker/provider.go
richardcase: Added #330

PR #344, opened 2022-01-05
Set ordered names for guest devices

### LINKED ISSUE #340: Set ordered names for guest devices
~We default the metadata device to `mds0`, yet firecracker seems to do what it wants and always creates it as `eth0`. Rude.~

~So we should either:~
- ~Find out why/where we are being ignored and stop it, or~
- ~Just give in and use set `eth0` (we would then have to ensure that none of the other network interfaces use `eth0`)~

~This is not causing a problem that we can see just yet, but it may in the future and it is confusing to debug.~

Update: this is by design https://weaveworks.slack.com/archives/C01SYJ779SR/p1641379724304500

yitsushi:

* We will rely on the order specified in the incoming grpc request and
  the fact that serialization in go shouldn't change the order
* We will update flintlock docs to say that if you care about the order
  of your network interfaces then you must specify them in the order
  from eth1,eth2,ethX
* MMDS device is the first (eth0)

> There is no relation between the numbering of the /network-interface
> API calls and the number of the network interface in the guest. Rather,
> it is usually the order of network interface creation that determines
> the number in the guest (but this depends on the distribution).
>
> For example, when you create two network interfaces by calling
> /network-interfaces/1 and then /network-interfaces/0, it may result in
> this mapping:
>
>    /network-interfaces/1 -> eth0
>    /network-interfaces/0 -> eth1

Source: https://github.com/firecracker-microvm/firecracker/blob/83ec483d4254936e1e2fd4adfe94761611a74a93/FAQ.md#how-does-network-interface-numbering-work

Fixes #340

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


api/types/microvm.proto
```diff
@@ -91,8 +93,9 @@ message NetworkInterface {
     // TAP represents a network interface that is a tap.
     TAP = 1;
   }
-  // GuestDeviceName is the name of the network interface to create in the microvm.
-  string guest_device_name = 1;
+  // DeviceID is the ID of the interface. There is no relation between the ID
+  // and the name of the iterface device on the quest machine.
```
Callisto13, line 97:
```suggestion
  // and the name of the interface device on the quest machine.
```

api/types/microvm.proto
```diff
@@ -46,6 +46,8 @@ message MicroVMSpec {
   repeated Volume additional_volumes = 9;
 
   // Interfaces specifies the network interfaces to be attached to the microvm.
+  // Device names on the guest machine are determines by the order defined in
```
Callisto13, line 49:
```suggestion
  // Device names on the guest machine are determined by the order defined in
```

Callisto13 (approved):
2 tiny typos but otherwise lgtm

Callisto13 (dismissed):

Callisto13:
actually you'll probably also need to regenerate something after changing those comments

Callisto13 (approved):

Issue #343, opened 2022-01-05
(provision): Bind flintlock grpc to private ip on start


Callisto13:
Update the service file and provision.sh to bind the grpc server to the device's local ip (perhaps with a flag to customise)
Currently the server starts on 0.0.0.0, which is not ideal since we have no auth at the moment.
Since the macvtap device address needs to be connected to by capmvm from within the same network, it makes sense to have users create their management cluster in a shared vlan. So it actually solves to problems to have the service start flintlockd on the local ip.

The user docs will also need to be updated to instruct on this.

PR #345, opened 2022-01-05
feat: bind flintlock server to internal ip

Callisto13:
This is for the provisioning script. There will be docs added.

There is an option to not do this, but by default the provision script
will assume you want a VLAN type setup.

There is a hideous little func which is sedding the flintlock service
file, and the service file itself does not look to pretty, but these are
temporary measures just for now.

Closes https://github.com/weaveworks/flintlock/issues/343

There is also a small change which will parse the lvm disk and grab the name, just in case people don't read the docs.

### DIFF
--- a/hack/scripts/provision.sh
+++ b/hack/scripts/provision.sh
@@ -270,8 +270,9 @@ install_firecracker() {
 #
 do_all_flintlock() {
 	local version="$1"
+	local address="$2"
 	install_flintlockd "$version"
-	start_flintlockd_service
+	start_flintlockd_service "$address"
 }
 
 # Fetch and install the flintlockd binary at the specified version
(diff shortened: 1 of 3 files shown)


hack/scripts/provision.sh
```diff
@@ -294,20 +295,38 @@ install_flintlockd() {
...
 	parent=$(ip route show | awk '/default/ {print $5}')
 	sed -i "s/PARENT_IFACE/$parent/" "$FLINTLOCKD_SERVICE_FILE"
 	socket="$CONTAINERD_STATE_DIR/containerd.sock"
 	sed -i "s|\(socket=\).*\(\\)|\1$socket \\\\\2|g" "$FLINTLOCKD_SERVICE_FILE"
 
-	start_service "$FLINTLOCK_BIN"
+	if [[ -z "$address" ]]; then
+		address=$(lookup_address)
+	fi
+	sed -i "s/ADDRESS/$address/" "$FLINTLOCKD_SERVICE_FILE"
```
yitsushi, line 324: I think that `-i` can cause trouble, if it does not, it will.

yitsushi (approved):
That `-i` is sus, but other than that lgtm

hack/scripts/provision.sh
Callisto13: yeh, a lot of those `-i`s are a prblem, but hopefully i can remove that whole bit soon

yitsushi (approved):

PR #394, opened 2022-02-15
Set instance_id in flintlock

richardcase:

This change will set the `instance_id` in the metadata if its not set to the UID for the microvm

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/core/application/app_test.go
+++ b/core/application/app_test.go
@@ -2,6 +2,7 @@ package application_test
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"testing"
 	"time"
@@ -9,8 +10,10 @@ import (
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/gomega"
 	"github.com/spf13/afero"
+	"sigs.k8s.io/yaml"
 
 	"github.com/weaveworks/flintlock/api/events"
+	"github.com/weaveworks/flintlock/client/cloudinit/instance"
 	"github.com/weaveworks/flintlock/core/application"
 	"github.com/weaveworks/flintlock/core/models"
 	"github.com/weaveworks/flintlock/core/ports"
@@ -55,7 +58,7 @@ func TestApp_CreateMicroVM(t *testing.T) {
 					}),
 				).Return(nil, nil)
 
-				expectedCreatedSpec := createTestSpec("id1234", defaults.MicroVMNamespace, testUID)
+				expectedCreatedSpec := createTestSpecWithMetadata("id1234", defaults.MicroVMNamespace, testUID, createInstanceMetadatadata(t, testUID))
 				expectedCreatedSpec.Spec.CreatedAt = frozenTime().Unix()
 				expectedCreatedSpec.Status.State = models.PendingState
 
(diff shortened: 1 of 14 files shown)


core/application/app_test.go
```diff
@@ -2,15 +2,18 @@ package application_test
...
 	"context"
+	"encoding/base64"
 	"errors"
 	"testing"
 	"time"
 
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/gomega"
 	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
```
bigkevmcd, line 13: I'd probably use https://pkg.go.dev/sigs.k8s.io/yaml instead of v2 of go-yaml

core/application/commands.go
```diff
@@ -60,6 +65,10 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
 		}
 	}
 
+	if addErr := a.addInstanceData(mvm, logger); addErr != nil {
```
bigkevmcd, line 68: I'm not sure there's any benefit in `addErr` here?

You've already returned if `err` wasn't nil earlier, which implies you don't want to retain the value ?

core/application/commands.go
```diff
@@ -122,3 +131,40 @@ func (a *app) DeleteMicroVM(ctx context.Context, uid string) error {
 
 	return nil
 }
+
+func (a *app) addInstanceData(vm *models.MicroVM, logger *logrus.Entry) error {
+	instanceData := &cloudinit.Metadata{}
+
+	meta, ok := vm.Spec.Metadata["meta-data"]
+	if ok {
```
bigkevmcd, line 139: Could you drop the comma, ok mechanism here, and check for vm.Spec.Metadata not being ""?

You can decode `""` but it would fail the YAML decoding, which implies that `""` is not considered a valid value

core/application/commands.go
```diff
@@ -122,3 +131,40 @@ func (a *app) DeleteMicroVM(ctx context.Context, uid string) error {
...
+		logger.Info("Instance metadata exists")
```
bigkevmcd, line 140: Is there any extra logging data that could come out here? the MicroVM ID perhaps?

bigkevmcd (commented):
Looks good, nothing major.

core/application/commands.go
richardcase: That was a mistake but i have changed this to reuse `err`

core/application/app_test.go
richardcase: Changed to use this wrapper.

core/application/commands.go
richardcase: Changed the logic here based on your suggestion

core/application/commands.go
richardcase: Added.

Callisto13: > Added.

where?

infrastructure/grpc/convert.go
```diff
@@ -32,6 +32,7 @@ func convertMicroVMToModel(spec *types.MicroVMSpec) (*models.MicroVM, error) {
 			},
 			VCPU:       int64(spec.Vcpu),
 			MemoryInMb: int64(spec.MemoryInMb),
+			Metadata:   map[string]string{},
```
Callisto13, line 35: This could technically be an `instance.New()` i guess

Callisto13 (approved):
What's the plan for those `MetadataOption` funcs?

richardcase:
> What's the plan for those `MetadataOption` funcs?

I'm going to change CAPMVM to use them :)

yitsushi (approved):
It's funny to see my original `map` solution comes back and says "hi" :smile:

yitsushi (approved):

PR #461, opened 2022-06-08
WIP: Basic auth and TLS

richardcase:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #462, opened 2022-06-09
feat: Add (very) basic auth

### LINKED ISSUE #356: Implement simple AUTH
Add a very basic thing to use a token with the API. Does not need to be perfect yet.
The default should be to start the server with auth.
For bonus points, add a flag to start the server with auth disabled.

Or do the opposite of that. Whatever you like.

Callisto13:
This commit adds some very basic token-based auth to the flintlock
server.

```
flintlockd run --basic-auth-token foo
```

Note that this is not a bearer token, or a username+password combo. It
is simply a token which will be compared against an **encoded** token on
all server calls.

Clients are responsible for adding an `authorization: basic <encoded
token>` header to all calls.

Tested manually by fiddling with `hammertime`, looking to add some e2e
if possible.

TLS is next on the list.

Based upon #461 

Closes #356 

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 
 require (
 	github.com/gorilla/mux v1.8.0
+	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/urfave/cli/v2 v2.8.1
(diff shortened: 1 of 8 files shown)


richardcase (approved):

Callisto13:
o shit forgot about all this

PR #483, opened 2022-07-14
patch: Provision flintlock from latest release

Callisto13:
Previously flintlock only had pre-releases, which meant a different
install method in `provision.sh`. Now it can use the standard func.

Also update the compatibility table and add a note about supported
Firecracker versions.

### DIFF
--- a/hack/scripts/provision.sh
+++ b/hack/scripts/provision.sh
@@ -196,15 +196,6 @@ latest_release_tag() {
 	echo "$url"
 }
 
-# Returns the tag associated with a "latest" pre-release (pre-releases do not show
-# up in the API as used in the above latest_release_tag func)
-latest_pre_release_tag() {
-	local repo_name="$1"
-	tag=$(git ls-remote --tags --sort="v:refname" "https://github.com/$repo_name" |
-		tail -n 1 | sed 's/.*\///; s/\^{}//')
-	echo "$tag"
-}
-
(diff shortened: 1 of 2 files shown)


hack/scripts/provision.sh
```diff
@@ -294,7 +294,7 @@ install_flintlockd() {
 	say "Installing flintlockd version $tag to $INSTALL_PATH"
 
 	if [[ "$tag" == "$DEFAULT_VERSION" ]]; then
-		tag=$(latest_pre_release_tag "$FLINTLOCK_REPO")
+		tag=$(latest_release_tag "$FLINTLOCK_REPO")
```
yitsushi, line 297: Do we need the `latest_pre_release_tag` function after this patch?

yitsushi (dismissed):

hack/scripts/provision.sh
Callisto13: technically no. i know i am usually ruthless when it comes to deleting dead code, but for some reason I had great attachment to this one 🤣

Callisto13: it's dead i killed it hope you are happy 😭 😭

richardcase (approved):

hack/scripts/provision.sh
yitsushi: :scream:

PR #484, opened 2022-07-18
feat: Add "insecure" opt to provisioner

Callisto13:
This will give the option to start flintlock without basic auth or
certs.
I am adding the options for those next, just needed `--insecure` first.

### DIFF
--- a/hack/scripts/provision.sh
+++ b/hack/scripts/provision.sh
@@ -273,6 +273,7 @@ do_all_flintlock() {
 	local version="$1"
 	local address="${2%:*}"
 	local parent_iface="$3"
+	local insecure="$4"
 
 	install_flintlockd "$version"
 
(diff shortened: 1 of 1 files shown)


hack/scripts/provision.sh
```diff
@@ -321,6 +323,7 @@ containerd-socket: "$CONTAINERD_STATE_DIR/containerd.sock"
 grpc-endpoint: "$address:9090"
 verbosity: 9
 parent-iface: "$parent_iface"
+insecure: $insecure
```
yitsushi, line 326: Any checks if `$insecure` contains `1|0` or `true|false`, or in that case we just say "user error, Read The Fancy Manual"

yitsushi (approved):

hack/scripts/provision.sh
Callisto13: I am doing more around this soon, so this will likely change, just need to get unblocked on something.

We don't validate other inputs (for this users cannot actually give any input since it is a bool flag) so perhaps that is a ticket for another day

PR #495, opened 2022-08-04
Update quickstart firecracker notes

Callisto13:
We have a released binary now so we don't need to tell people to build it.


richardcase (approved):
Looks good to me :)

PR #514, opened 2022-08-19
refactor: move validation/defaults to app

### LINKED ISSUE #513: Move defaulting and validation from gRPC to App

Currently we do some [defaulting](https://github.com/weaveworks-liquidmetal/flintlock/blob/main/infrastructure/grpc/defaults.go) and [validation](https://github.com/weaveworks-liquidmetal/flintlock/blob/main/infrastructure/grpc/server.go#L58) withinthe gRPC infrastructure implementation. This is an application level concern and should live within the core "application".  

This means if we decide to expose a graphQL [insert your favourite api type] then we still get validation/defaulting.

richardcase:

This change moves the adding of the mmds network interface and
validation of the microvm spec to the app and out of the gRPC
infrastructure.

Fixes #513 

**Checklist**:


core/application/commands.go
```diff
@@ -171,3 +186,24 @@ func (a *app) addInstanceData(vm *models.MicroVM, logger *logrus.Entry) error {
 
 	return nil
 }
+
+func (a *app) addMetadataInterface(mvm *models.MicroVM) error {
```
Callisto13, line 190: This could probably return a `bool` rather than an `error`. The caller is doing `fmt.Errorf("adding metadata network interface: %w", err)` but the func never returns an err.
Actually, in that case does it need to return anything? It could just `return` nothing on 194 and leave the func early

richardcase: Thats a good point. Changed.

core/application/commands.go
```diff
@@ -171,3 +183,24 @@ func (a *app) addInstanceData(vm *models.MicroVM, logger *logrus.Entry) error {
...
+		GuestDeviceName:       MetadataInterfaceName,
+		Type:                  models.IfaceTypeTap,
+		AllowMetadataRequests: true,
+		GuestMAC:              "AA:FF:00:00:00:01",
+		StaticAddress: &models.StaticAddress{
+			Address: "169.254.0.1/16",
+		},
+	})
+
+	return
```
Callisto13, line 205: the return is likely redundant

Callisto13 (approved):

Issue #417, opened 2022-04-22
Add support for Cloud Hypervisor


richardcase:
**Describe the solution you'd like:**
I would like the option to use Cloud Hypervisor as the microvm implementation as an alternative to Firecracker. There are reasons why i might want to do this, for example:

- GPU support
- Enclave support (i.e. SGX)

**Why do you want this feature:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #515, opened 2022-08-19
feat: add new disk service

richardcase:

This change adds a new port for a "disk service" that will be used to
create disk images (i.e. img/iso files). An infrastructure
implementation has been added using godisk.

Relates #417 

Note: this isn't used by anything in this change. But it will be used in
the future with the Cloud Hypervisor implementation.

**Checklist**:

- [x] adds unit tests


core/ports/services.go
```diff
@@ -145,3 +145,42 @@ type DeleteIfaceInput struct {
...
+	// DiskTypeISO9660 is an iso filesystem.
+	DiskTypeISO9660
+)
+
+// DiskCreateInput are the input options for creating a disk.
+type DiskCreateInput struct {
+	//Path is the filesystem path of where to create the disk.
+	Path string
+	// Size is how big the disk should be. It uses human readable formats
+	// suck as 8Mb, 10Kb.
```
Callisto13, line 170:
```suggestion
	// such as 8Mb, 10Kb.
```

infrastructure/godisk/disk_service.go
```diff
@@ -0,0 +1,93 @@
...
+		return errPathRequired
+	}
+	if input.Size == "" {
+		return errSizeRequired
+	}
+
+	if _, err := s.fs.Stat(input.Path); err == nil {
+		if removeErr := s.fs.Remove(input.Path); removeErr != nil {
+			return fmt.Errorf("removing disk %s: %w", input.Path, removeErr)
+		}
```
Callisto13, line 41: can you explain this bit here: why do we delete whatever it at the input path first? it is to clean whatever is there before we create over it?

infrastructure/godisk/disk_service.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return fmt.Errorf("converting disk size %s: %w", input.Size, err)
+	}
+
+	createdDisk, err := diskfs.Create(input.Path, int64(diskSize), diskfs.Raw)
+	if err != nil {
+		return fmt.Errorf("creating disk %s: %w", input.Path, err)
+	}
+
+	createdDisk.LogicalBlocksize = 512
```
Callisto13, line 56: can we have that in a const?

infrastructure/godisk/disk_service.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = rw.Write(decoded)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
Callisto13, line 92: could be `return err` but maybe that would read weird

internal/inject/wire.go
```diff
@@ -87,7 +89,7 @@ func appConfig(cfg *config.Config) *application.Config {
 	}
 }
 
-func appPorts(repo ports.MicroVMRepository, prov ports.MicroVMService, es ports.EventService, is ports.IDService, ns ports.NetworkService, ims ports.ImageService, fs afero.Fs) *ports.Collection {
+func appPorts(repo ports.MicroVMRepository, prov ports.MicroVMService, es ports.EventService, is ports.IDService, ns ports.NetworkService, ims ports.ImageService, fs afero.Fs, ds ports.DiskService) *ports.Collection {
```
Callisto13, line 92: hrmmm could do with a way of shrinking this signature

core/ports/services.go
richardcase: My code sucks :rofl:

richardcase: Changed.

infrastructure/godisk/disk_service.go
richardcase: Yep it's basically that. If the disk image exists already then delete and we will re-create.

What do you think to that default behaviour? Should i make it configurable? Perhaps add an option to `DiskCreateInput` valled `OverwriteExisting`?

infrastructure/godisk/disk_service.go
richardcase: Yep, that would've been a const to start with :+1:

infrastructure/godisk/disk_service.go
richardcase: I like it, a lot neater.

internal/inject/wire.go
richardcase: I did link about that but thought as this is for DI i wouldn't bother....it can get a bit messy with wire.

infrastructure/godisk/disk_service.go
Callisto13: maybe.

worst case is someone sets Path to something they should not and it is erased. having it error unless told explicitly to overwrite may be useful

richardcase: Agreed, changing now...

richardcase: @Callisto13 - all changed.

Callisto13 (approved):
LGTM

PR #519, opened 2022-08-22
feat: force dhcp to use mac as client identifier

### LINKED ISSUE #518: Force DHCP to use mac address

Currently when creating the cloud-init network configuration we don't force DHCP to use the mac address as the client identifier. This can cause a dhcp server to not assign an ip address to a microvm.

richardcase:

This change will force DHCP to use the mac address as the client
identifier when requesting a mac address. Without this (depending on
your distro) it may not use a mac address and the mv may not get a ip
address assigned to it.

Signed-off-by: Richard Case <richard.case@outlook.com>

Fixes #518 

**Checklist**:

### DIFF
--- a/infrastructure/firecracker/config.go
+++ b/infrastructure/firecracker/config.go
@@ -217,9 +217,10 @@ func generateNetworkConfig(vm *models.MicroVM) (string, error) {
 		macAddress := getMacAddress(&iface, status)
 
 		eth := &cinetwork.Ethernet{
-			Match: cinetwork.Match{},
-			DHCP4: firecracker.Bool(true),
-			DHCP6: firecracker.Bool(true),
+			Match:          cinetwork.Match{},
+			DHCP4:          firecracker.Bool(true),
+			DHCP6:          firecracker.Bool(true),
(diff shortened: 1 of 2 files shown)


infrastructure/firecracker/config.go
```diff
@@ -217,9 +217,10 @@ func generateNetworkConfig(vm *models.MicroVM) (string, error) {
 		macAddress := getMacAddress(&iface, status)
 
 		eth := &cinetwork.Ethernet{
-			Match: cinetwork.Match{},
-			DHCP4: firecracker.Bool(true),
-			DHCP6: firecracker.Bool(true),
+			Match:          cinetwork.Match{},
+			DHCP4:          firecracker.Bool(true),
+			DHCP6:          firecracker.Bool(true),
+			DHCPIdentifier: firecracker.String("mac"),
```
Callisto13, line 223: would be good to have that as a const, but otherwise LGTM

Callisto13 (approved):

infrastructure/firecracker/config.go
richardcase: Good idea, changed.

Callisto13 (approved):

PR #543, opened 2022-09-28
Add --bridge flag to flintlock provision script

Callisto13:
The script flags for flintlock are getting a bit out of hand, I have a plan to deal with this but won't have time to get to it for a bit, so this will have to do.


hack/scripts/provision.sh
```diff
@@ -313,16 +316,25 @@ write_flintlockd_config() {
 containerd-socket: "$CONTAINERD_STATE_DIR/containerd.sock"
 grpc-endpoint: "$address:9090"
 verbosity: 9
-parent-iface: "$parent_iface"
 insecure: $insecure
 EOF
 
+	if [[ -n "$bridge_name" ]]; then
```
richardcase, line 322: You can supply both bridge name and parent iface to flintlock

hack/scripts/provision.sh
```diff
@@ -966,6 +990,7 @@ cmd_flintlock_help() {
       --version, -v      Version to install (default: latest)
       --grpc-address, -a Address on which to start the GRPC server (default: local ipv4 address)
       --parent-iface, -i Interface of the default route of the host
+      --bridge, -b       Bridge to use instead of an interface (will override --parent-iface)
```
richardcase, line 993: Same comment as above

hack/scripts/provision.sh
richardcase: Discussed on slack

richardcase (approved):

PR #549, opened 2022-10-10
fix: issue systemctl commands to the correct containerd service

ianbuss:
**What this PR does / why we need it**:

This is a small fix to change the name of the containerd systemd service to `containerd-dev.service` when deployed with `--dev`. Previously systemctl commands in the script were issued to the `containerd.service`.

**Which issue(s) this PR fixes**:
Fixes #548 

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits into logical changes
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #552, opened 2022-10-11
feat: Update provision.sh to detect arch

### LINKED ISSUE #376: Update provision.sh to install arm64 bins
The provision script defaults to amd64, we should have an arm option. amd should be the default

Example commands:

```
./hack/scripts/provision.sh flintlock --arch <amd64|arm>

./hack/scripts/provision.sh containerd --arch <amd64|arm>

./hack/scripts/provision.sh firecracker --arch <amd64|arm>

./hack/scripts/provision.sh all --arch <amd64|arm>
```

Containerd and flintlock already ship arm bins, we will need to play https://github.com/weaveworks/flintlock/issues/375 first

### LINKED ISSUE #548: Provision script cannot download arch-specific firecracker builds
**What happened:**
With the new release of the macvtap fork of firecracker, the binaries are now `firecracker_amd64` and `firecracker_arm64` rather than just `firecracker`. The provision script fails to download the latest release.

The provision script should download the right binary and place it in `/usr/local/bin/firecracker`.

**How to reproduce it:**
Run `./provision.sh firecracker` from the `hack/scripts` dir.

**Environment:**

- flintlock version: HEAD of main

Callisto13:
edit: Actually realised I don't like the flags and that the idea in #549 is correct: we should detect.

Also take care of any backwards compatibility for those downloading earlier firecracker versions which were named differently.

Closes #376 
Closes #548

### DIFF
--- a/hack/scripts/provision.sh
+++ b/hack/scripts/provision.sh
@@ -13,6 +13,7 @@ OPT_UNATTENDED=false
 DEVELOPMENT=false
 DEFAULT_VERSION="latest"
 DEFAULT_BRANCH="main"
+ARCH=amd64
 
 # paths to be set later, put here to be explicit
 CONTAINERD_CONFIG_PATH=""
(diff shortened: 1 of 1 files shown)


Callisto13:
Reminder to make FC release a non-pre release after this is merged

richardcase (approved):

PR #575, opened 2022-10-26
Revert "test: use our ghcr images in e2es (#572)"

Callisto13:
I realised that using these images requires some more networking stuff than I have patience to set up in tests.


yitsushi (approved):
What networking stuff? Just to change the image? :O

Callisto13:
mvms with the "official" images get all mad when they can't get an ip assigned

PR #583, opened 2022-11-07
Rename docs deploy key

Callisto13:
I don't know why github didn't want to see the old one, even when I
replaced it.
Now with the same key under a new name it is fine so.... yeh.

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -33,7 +33,7 @@ jobs:
           go-version: '1.18'
       - uses: webfactory/ssh-agent@v0.5.0
         with:
-          ssh-private-key: ${{ secrets.FLINTLOCK_DOCS_WEAVEWORKS_DOCS_BOT_DEPLOY_KEY }}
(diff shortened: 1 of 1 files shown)


Skarlso:
Can you help me understand how rename docs deploy key is a workflow trigger change? 🤔

yitsushi (approved):

Callisto13:
> Can you help me understand how rename docs deploy key is a workflow trigger change?

It doesn't change the trigger, just fixes some issue I can't explain.

Github started failing with this: https://github.com/weaveworks-liquidmetal/flintlock/actions/runs/3386049291/jobs/5625278172

I replaced the key it complained about, but it still errored. So then I just created a new key, with the same values as the replacement and it works so i am taking it as a win and leaving it.

Skarlso:
Ah gotcha. Thanks for the context. :)

Callisto13:
note that the flintlock docs have vanished from the domain becuase this is the first time we have published them since we moved the repos to the new org and i need to do some work to verify the domain

PR #614, opened 2023-01-13
Temporarily disable pre-release tests

Callisto13:
The e2es have been off for a while since I lost auth powers to create new projects in Equinix. Since then something weird has happened and the tests fail strangely for different reasons.

I have performed all the test steps manually several times (all they do is simple CRUD operations on flintlock) and the code seems fine. I have also used this version of flintlock to create a few capmvm clusters. So I am confident that the problem is with the tests and not the code.

I am disabling the test step of the release process so that I can release flintlock and capmvm (thus close off a large image-builder epic) before I circle back to figure out what is going on with these tests.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,23 +6,23 @@ on:
     - "v*.*.*"
 
 jobs:
-  test:
-    runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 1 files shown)


yitsushi (approved):

yitsushi:
:crossed_fingers:

PR #627, opened 2023-02-11
refactor: removed `gw` command

richardcase:

The separate `gw` command has been removed and its functionality incorporated into the `run` command. The HTTP gateway is then enabled using `--enable-http`.

NOTE: the code to start the HTTP gateway remains the same. We will need to refactor it in the future to take into account the the TLS flags.

The reason for making this change is that we should be able to enable the grpc and http endpoints with the same instance of flintlock instead of having to start 2 instances.

Fixes #

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


Callisto13 (dismissed):

richardcase:
As this was previously approved by @Callisto13 and its only a rebase, merging

Callisto13 (approved):

yitsushi (approved):

Callisto13:
ummm i don't think this was ready to go @yitsushi 

I remember @richardcase saying something about forgetting to do auth and other bits

yitsushi:
Well it was not in Draft mode and not much changes since you approved 2 weeks ago.

Callisto13:
yeh this was a verbal chat in the community meeting

richardcase:
It was ready to go. This is the note I added to the PR:

> NOTE: the code to start the HTTP gateway remains the same. We will need to refactor it in the future to take into account the TLS flags.

So, this was refactoring and not adding TLS etc, which would need to be done separatly.

PR #664, opened 2023-03-10
test: Update network-interface config for e2es

Callisto13:
This was previously `eth0` which led to a weird and misleading error from firecracker complaining about `network_interfaces` missing.. when the api actually accepts `network-interfaces` with a dash. This is a red-herring: really firecracker just wasn't happy with us setting `eth0` because it apparently uses that for something else.

I also updated the images used in the test to use our official builds.

### DIFF
--- a/test/e2e/utils/utils.go
+++ b/test/e2e/utils/utils.go
@@ -79,8 +79,8 @@ func PidRunning(pid int) bool {
 
 func defaultTestMicroVM(name, namespace string) *types.MicroVMSpec {
 	var (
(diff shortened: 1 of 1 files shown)


yitsushi (approved):
:shipit: 

Can you document this something, somewhere? Or is it's not expected users will end up setting it to eth0?

Callisto13:
Yeh i will do a separate PR with a note and adding the failure to the troubleshooting guides, because it is the most irritating error

PR #718, opened 2023-05-25
Remove `Volume.size_in_mb` from api; improve `pkg/queue` tests

### LINKED ISSUE #553: Setting root_volume.size_in_mb seems to have no effect
It would appear we don't actually process that input and resize the volume before mountind.

cf: https://github.com/weaveworks-liquidmetal/cluster-api-provider-microvm/issues/234

aryan9600:
kind/cleanup

Remove `Volume.size_in_mb` since its unused and its non trivial to actually use it for its intended purpose.
Furthermore, remove `t.Fatal` calls from non test goroutines. Currently running `go vet ./...` would return:
```
pkg/queue/queue_test.go:50:6: call to (*T).Fatal from a non-test goroutine
pkg/queue/queue_test.go:94:6: call to (*T).Fatal from a non-test goroutine
```

Fixes #553 

**Checklist**:

- [ ] includes documentation
- [ ] adds unit tests


api/types/microvm.proto
```diff
@@ -142,8 +142,6 @@ message Volume {
   VolumeSource source = 4;
   // PartitionID is the uuid of the boot partition.
   optional string partition_id = 5;
-  // Size is the size to resize this volume to.
```
Skarlso, line 145: I know it's an optional field, but shouldn't we be deprecating this first?

richardcase: We'd need to add `reserved 6`  and probably `reserved "size_in_mb"` to the definition of Volume.

PR #995, opened 2024-12-27
[DRAFT WORK IN PROGRESS] Adding VirtioFS

steve-fraser:
Features adding VirtioFS Support

In a upcoming PR I will be adding

virtiofs_create_test.go
virtiofs_create_delete_test.go

These tests will be testing the create and delete functionality to make sure it's valid on what we expect them to be


core/application/commands.go
```diff
@@ -37,6 +37,29 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
 		return nil, fmt.Errorf("an error occurred when attempting to validate microvm spec: %w", validErr)
 	}
 
+	if mvm.Spec.RootVolume.Source.VirtioFS != nil {
+		return nil, fmt.Errorf("VirtioFS is not available yet for RootVolume")
+	}
+	if len(mvm.Spec.AdditionalVolumes) > 0 {
```
richardcase: Its probably worth checking as well that if `volume.Source.VirtioFS` is specified that the microvm provider has the capability.

core/application/commands.go
```diff
@@ -37,6 +37,29 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
 		return nil, fmt.Errorf("an error occurred when attempting to validate microvm spec: %w", validErr)
 	}
 
+	if mvm.Spec.RootVolume.Source.VirtioFS != nil {
```
richardcase: I would move this to a "custom" validator on the struct.  For example:

```go
	// RootVolume specified the root volume to be attached to the machine.
	RootVolume Volume `json:"root_volume" validate:"required,novirtiofs"`
```
([source](https://github.com/liquidmetal-dev/flintlock/blob/main/core/models/microvm.go#L41))

And then register the tag and add the validation function here https://github.com/liquidmetal-dev/flintlock/blob/main/pkg/validation/validate.go

core/application/commands.go
```diff
@@ -37,6 +37,29 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
...
+		return nil, fmt.Errorf("VirtioFS is not available yet for RootVolume")
+	}
+	if len(mvm.Spec.AdditionalVolumes) > 0 {
+		virtioFSCount := 0 // Counter for VirtioFS volumes
+		
+		for _, volume := range mvm.Spec.AdditionalVolumes {
+			if volume.Source.Container != nil && volume.Source.VirtioFS != nil {
```
richardcase: Like above we could add a custom validator and attach it to the source field: https://github.com/liquidmetal-dev/flintlock/blob/main/core/models/volumes.go#L10

core/application/commands.go
```diff
@@ -37,6 +37,29 @@ func (a *app) CreateMicroVM(ctx context.Context, mvm *models.MicroVM) (*models.M
...
+			}
+			
+			// Check if this volume has VirtioFS
+			if volume.Source.VirtioFS != nil {
+				virtioFSCount++
+			}
+		}
+	
+		// Ensure only one volume contains VirtioFS
+		if virtioFSCount > 1 {
```
richardcase: We could add this check to the "additional volumes" as a custom validator as well.

core/plans/microvm_create_update.go
```diff
@@ -66,6 +66,13 @@ func (p *microvmCreateOrUpdatePlan) Create(ctx context.Context) ([]planner.Proce
 		return nil, fmt.Errorf("adding root dir step: %w", err)
 	}
 
+	// MicroVM provider doesn't have virtiofs
+	if provider.Capabilities().Has(models.VirtioFSCapability) {
```
richardcase, line 70: We need to make sure that we have the earlier validation check to fail early when the command first executes.

core/steps/runtime/virtiofs.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return fmt.Errorf("opening sterr file %s: %w", virtiofsState.VirtioFSStderrPath(), err)
+	}
+
+	cmdVirtioFS.Stderr = stdErrFileVirtioFS
+	cmdVirtioFS.Stdout = stdOutFileVirtioFS
+	cmdVirtioFS.Stdin = &bytes.Buffer{}
+
+	var startErr error
+	process.DetachedStart(cmdVirtioFS)
```
richardcase: If we start `virtiofsd` here, we'll need to store the PID somewhere so that when we delete the microvm we know to stop the process.

richardcase: Otherwise we'll get orphaned processes.

infrastructure/microvm/cloudhypervisor/create.go
```diff
@@ -110,7 +110,7 @@ func (p *provider) buildArgs(vm *models.MicroVM, state State, _ *logrus.Entry) (
 
 	// CPU and memory
 	args = append(args, "--cpus", fmt.Sprintf("boot=%d", vm.Spec.VCPU))
-	args = append(args, "--memory", fmt.Sprintf("size=%dM", vm.Spec.MemoryInMb))
+	args = append(args, "--memory", fmt.Sprintf("size=%dM,shared=on", vm.Spec.MemoryInMb))
```
richardcase: Is `shared=on` required only for virtiofs?

infrastructure/microvm/cloudhypervisor/create.go
```diff
@@ -119,6 +119,8 @@ func (p *provider) buildArgs(vm *models.MicroVM, state State, _ *logrus.Entry) (
 	}
 	args = append(args, "--disk", "path="+rootVolumeStatus.Mount.Source)
 	args = append(args, fmt.Sprintf("path=%s,readonly=on", state.CloudInitImage()))
+	args = append(args, "--fs", fmt.Sprintf("tag=user,socket=%s,num_queues=1,queue_size=1024", state.VirtioFSPath()))
```
richardcase: Guessing we only need this if we are using virtiosfs. We should probably have a check and only add this arg if required.

core/steps/runtime/virtiofs.go
richardcase: (i see there is the file path on the state....just need to use that somewhere).

infrastructure/virtiofs/virtiofs.go
```diff
@@ -0,0 +1,45 @@
+package virtiofs
```
richardcase, line 1: The "infrastructure" package is normally used to provide a concrete implementation of a service (a.k.a an **adapter**) that is attached via a **port** to the core application. This is from hexagonal architecture: https://medium.com/ssense-tech/hexagonal-architecture-there-are-always-two-sides-to-every-story-bc0780ed7d9c

For example, there is the [MicroVMService](https://github.com/liquidmetal-dev/flintlock/blob/main/core/ports/services.go#L12) port and the core application uses this. It not aware of the actual infrastructure. We then have to 2 adapters (i.e. Firecracker and CloudHypervisor) that can plug into this port, [both of which are defined](https://github.com/liquidmetal-dev/flintlock/tree/main/infrastructure/microvm) in the infrastructure package.

I'm wondering if there is a way that we can split this up into port & adapter? Would we want to mock virtiofs at all?

richardcase (commented):
Great work @steve-fraser 🎉 Please don't be put off by the number of comments. Let me know if i can help with anything.

core/application/commands.go
steve-fraser: Can you check if the most recent code update resolves this?

core/application/commands.go
steve-fraser: Can you check if the most recent code update resolves this?

core/application/commands.go
steve-fraser: Can you check if the most recent code update resolves this?

core/application/commands.go
steve-fraser: I'll look into this one still work in progress

core/models/microvm.go
```diff
@@ -38,9 +38,9 @@ type MicroVMSpec struct {
 	// NetworkInterfaces specifies the network interfaces attached to the machine.
 	NetworkInterfaces []NetworkInterface `json:"network_interfaces" validate:"required,dive,required"`
 	// RootVolume specified the root volume to be attached to the machine.
-	RootVolume Volume `json:"root_volume" validate:"required"`
+	RootVolume Volume `json:"root_volume" validate:"required,novirtiofs"`
```
richardcase, line 41: These struct level validations are great :+1:

core/models/volumes.go
```diff
@@ -55,6 +58,13 @@ type ContainerVolumeSource struct {
 	Image ContainerImage `json:"image"`
 }
 
+// VirtioFSSource represents the details of the VirtioFS volume.
+type VirtioFSVolumeSource struct {
+	// Image is the OCI image to use.
+	Path string `json:"path"`
```
richardcase, line 64: I was chatting with someone the other day about flintlock and they really wanted the equivalant of "host path" storage, so this is great.

core/models/volumes.go
```diff
@@ -55,6 +58,13 @@ type ContainerVolumeSource struct {
 	Image ContainerImage `json:"image"`
 }
 
+// VirtioFSSource represents the details of the VirtioFS volume.
+type VirtioFSVolumeSource struct {
```
richardcase, line 62: Do you think it would be good to add validation to the path to check it exists?

core/ports/services.go
```diff
@@ -187,3 +187,17 @@ type DiskFile struct {
...
+
+
+
+// Create VirtioFSInput are the input options for creating a disk.
+type VirtioFSCreateInput struct {
+	Path string
+}
+
+
+// MicroVMService is the port definition for a microvm service.
```
richardcase:
```suggestion
// MicroVMService is the port definition fora virtiofs service.
```

core/ports/services.go
```diff
@@ -187,3 +187,17 @@ type DiskFile struct {
...
+type VirtioFSService interface {
+	// Create will create a new microvm.
```
richardcase:
```suggestion
	// Create will create a new virtiofs share.
```

richardcase (commented):
This is looking great. I think having the port and adapter is a great addition.

core/ports/services.go
```diff
@@ -187,3 +187,23 @@ type DiskFile struct {
...
+}
+
+type VirtioFSDeleteInput struct {
+	Path string
+}
+
+// VirtiofsService is the port definition for a VirtioFS service.
+type VirtioFSService interface {
+	// Create will create a new virtiofs share.
+	Create(ctx context.Context, vmid *models.VMID, inout VirtioFSCreateInput) (*models.Mount,error)
```
richardcase:
```suggestion
	Create(ctx context.Context, vmid *models.VMID, input VirtioFSCreateInput) (*models.Mount,error)
```

richardcase: nit

richardcase (commented):
I think this looks good to go. Would you be able to do the following:

- confirm what testing you have done locally?
- describe any automated testing that we should add as a follow up to this PR
- Ideally, squash your commits so they tell a story. It doesn't have to be down to 1 but a smaller number is preferable to have a clean commit history.

PR #1016, opened 2025-02-04
fix release version for github action (issue: 556 action-gh-release)

steve-fraser:
Github release had a bad release version that was referenced.

https://github.com/softprops/action-gh-release/issues/556


richardcase:
Thats strange, dependabot should've kept this updated.

richardcase (approved):

PR #1049, opened 2025-03-24
adding pci devices support for flintlockd

steve-fraser:
Adding PCI devices support for Filntlockd


api/types/microvm.proto
```diff
@@ -171,6 +175,12 @@ message ContainerVolumeSource {
     string image = 1;
 }
 
+message PCIDevice {
+  string VendorID = 1;
+  string DeviceID = 2;
```
richardcase, line 180: We should probably model this for future expansion. For example, selecting a GPU for a pool of GPUs on the host.

Something along the lines of what we do with `VolumeSource`.

Maybe we have something like:

```proto
message PCIDevice {
  optional DeviceIDSource = 1;
}
```

Another option is that we model it on "requests". So the VM requests a PCI device that meets a set of criteria? And may the criteria is a CEL expression which would then cover a number of different options one being by DeviceID? I quite like this approach, its very similar to where Kubernetes is going

richardcase: If we used the CEL approach then it could end up being:

message PCIDevice {
  string Expression  = 1;
}

And then the expression would be something  like:

- `device.id == "58:00.0"`
- `device.class == "vga" && device.available`

richardcase: We could then also use CEL to select PCI devices available to flintlock.

steve-fraser: It's an interesting idea. How would you validate the cell expression?

richardcase: We'd need a custom data model for the CEL side and populate that with data from `/sys/bus/pci/devices` and other things.

richardcase: Maybe we could go with something like this to not block using the device id directly?:

```proto
message PCIDevice {
  optional DeviceIDSource = 1;
  optional Expression= 2; // Add this later
}
```

richardcase:
It would be good to get this in

Issue #204, opened 2021-11-04
Support snapshot/restore of microvms


richardcase:
**Describe the solution you'd like:**

Is there any need to support the snapshot create/restore functionality of Firecracker? This could support periodic backups or migrating a microvm from a host to a different host.

But...our core use case doesn't require this as machines as immutable.

**Why do you want this feature:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

richardcase:
Closing as its not needed for our core use case. Re-open is this required.

richardcase:
This is in progress

PR #1114, opened 2026-06-30
feat(snapshot): add microvm snapshot packaged as OCI image

richardcase:
## What

Adds a `SnapshotMicroVM` gRPC method that pauses a running microvm, captures a point-in-time snapshot, resumes it, and packages the snapshot artifacts into an **OCI image** written to an on-disk OCI layout. The image reference + digest are returned to the caller.

Implemented for **both firecracker and cloud-hypervisor** via a new `Snapshot` method on `MicroVMService`, gated by a new `SnapshotCapability`.

## Design

- **Synchronous** (not the reconciler path): snapshot is an imperative, point-in-time action with no persistent desired state, and the caller wants the resulting image back in the response — so it follows the direct query-style path, not the event/plan/step machinery.
- **Packaging is a new provider-agnostic `SnapshotPackager` port** (oras-go). The app orchestrates: `provider.Snapshot` (VMM pause→snapshot→resume, returns raw file paths) → `packager.Build` (oras → on-disk OCI layout at `<StateRootDir>/snapshots`, survives VM delete) → return ref+digest.
- **Self-describing image:** the microvm spec is included as a config blob alongside the snapshot layers, so a future restore can reconstruct the VM. Custom artifact type `application/vnd.flintlock.microvm.snapshot.v1`.
- **Failure semantics:** both providers pre-flight a running check and **always resume** via deferred resume, so the VM is never left paused on snapshot failure.

## Provider notes

- **cloud-hypervisor:** fixes the client `Snapshot` call to send the `destination_url` body (was a bodyless `PUT` — would never have worked).
- **firecracker:** enables the API socket at boot (`--no-api` → `--api-sock`) since firecracker snapshots are API-only.

## Tests

- packager (real oras layout: asserts artifact type, config blob, layer count, digest)
- CH client `destination_url` body guard (unix-socket server)
- gRPC handler validation + happy path
- app use-case capability-gate + happy path

`go build ./...` passes; `go test ./...` → 209 pass.

## Scope / follow-ups

Snapshot only. **Restore**, **remote registry push**, and firecracker **Diff** snapshots are out of scope (noted in code).

## Needs review attention

- **Not runtime-verified** — no live VMM available in the dev env. Needs an e2e run.
- The firecracker **`--api-sock` flip is unconditional** for all VMs. Please confirm the existing SIGHUP-based `Delete` and PID-only `State` still behave with the control socket present.
- **go.mod:** pinned oras `v2.5.0` + `x/sync v0.16.0` to keep the go directive at `1.23.0` (oras v2.6.x demands go 1.25).
Relates: #204


infrastructure/snapshot/packager.go
```diff
@@ -0,0 +1,122 @@
...
+	config *Config
+}
+
+// Build packages the snapshot artifacts and microvm spec into an OCI image
+// written to the on-disk layout and returns the resulting image details.
+func (p *packager) Build(ctx context.Context, input ports.SnapshotPackageInput) (*ports.SnapshotImage, error) {
+	if len(input.Artifacts) == 0 {
+		return nil, fmt.Errorf("no snapshot artifacts to package")
+	}
+
```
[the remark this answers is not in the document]
richardcase: Changed

pkg/cloudhypervisor/client.go
```diff
@@ -430,6 +430,7 @@ func (c *client) Snapshot(ctx context.Context, _ *VMSnapshotConfig) error {
 			405: "The VM instance could not be snapshotted because it is not booted",
 		})).
 		Put().
+		BodyJSON(config).
 		Fetch(ctx)
```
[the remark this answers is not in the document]
richardcase: Changed

core/application/snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	image, err := a.ports.SnapshotPackager.Build(ctx, ports.SnapshotPackageInput{
+		Reference: reference,
+		Artifacts: result.Artifacts,
+		Spec:      foundMvm,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("packaging snapshot into image: %w", err)
+	}
```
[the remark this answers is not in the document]
richardcase: Changed

richardcase (commented):
Review findings

infrastructure/microvm/firecracker/snapshot.go
```diff
@@ -0,0 +1,114 @@
...
+		SnapshotPath: &statePath,
+		MemFilePath:  &memPath,
+		SnapshotType: snapshotType,
+	}); err != nil {
+		return nil, fmt.Errorf("creating snapshot: %w", err)
+	}
+
+	logger.Info("snapshot taken")
+
+	return &ports.SnapshotResult{
```
richardcase, line 83: **[P1] Snapshot images omit writable disks.** This result packages only memory and VMM state, while the VM root and additional disks are writable external files. Restoring this state against fresh or subsequently modified disks will not reproduce the snapshot and may corrupt the guest. Please include snapshot-time disk contents, or reject writable volumes and explicitly require externally managed disk snapshots.

infrastructure/microvm/firecracker/snapshot.go
```diff
@@ -0,0 +1,114 @@
...
+	running, err := p.isRunning(vmState)
+	if err != nil {
+		return nil, err
+	}
+
+	if !running {
+		return nil, fmt.Errorf("cannot snapshot: %w", cerrs.ErrNotRunning)
+	}
+
+	scratch := vmState.Root() + "/snapshot"
```
richardcase, line 46: **[P1] Concurrent requests can corrupt snapshots.** Every request for this VM writes to the same fixed directory and filenames. Two requests can both pass the running check, interleave pause/snapshot/resume operations, overwrite each other, and have one request remove the directory while the other packages it. Please serialize snapshots per VM and/or allocate a unique scratch directory for each operation. The Cloud Hypervisor implementation has the same fixed-directory issue.

core/application/snapshot.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if !provider.Capabilities().Has(models.SnapshotCapability) {
+		return nil, coreerrs.NewNotSupported("snapshot")
+	}
+
+	logger = logger.WithField("vmid", foundMvm.ID)
+	logger.Infof("taking snapshot of microvm into %s", reference)
+
+	result, err := provider.Snapshot(ctx, ports.SnapshotInput{VMID: foundMvm.ID})
```
richardcase, line 50: **[P2] Failed snapshots leave scratch artifacts behind.** Cleanup is registered only after `provider.Snapshot` returns successfully. A provider error after creating or partially writing its fixed scratch directory, including a resume failure, therefore leaks potentially large snapshot files. Providers should clean partial output on every error, while preserving successful files until packaging completes. Firecracker documents that existing output files are truncated and overwritten, so stale files do not by themselves prevent a retry: https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#creating-full-snapshots

infrastructure/snapshot/packager.go
```diff
@@ -0,0 +1,152 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("packing snapshot manifest: %w", err)
+	}
+
+	if err := fileStore.Tag(ctx, manifestDesc, input.Reference); err != nil {
+		return nil, fmt.Errorf("tagging snapshot manifest: %w", err)
+	}
+
+	if err := os.MkdirAll(p.config.SnapshotRoot, 0o755); err != nil {
```
richardcase, line 128: **[P1] Enforce snapshot storage quotas and capacity checks.** This endpoint can persist a full guest-memory image on every request, and packaging temporarily retains both the raw snapshot and its OCI copy. There is no quota, reservation, retention limit, or free-space preflight before writing to the shared state filesystem, so an ordinary caller can exhaust host storage and disrupt the daemon. Firecracker explicitly requires integrators exposing snapshot triggers to enforce quotas against this DoS risk: https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#provisioning-host-disk-space-for-snapshots

richardcase (commented):
Additional findings from comparison with the [Firecracker snapshotting documentation](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md):

Beyond the inline findings, the public documentation should describe the restore-time contract: disk backing files and host resources must be recreated at compatible paths; snapshot memory must remain immutable while a restored VM uses it; network and vsock connections may not survive; MMDS data is not contained in Firecracker snapshot state; guest wall-clock correction may be required; and restore targets must satisfy Firecracker software and hardware compatibility requirements. At present only the generated RPC schema is added, so callers could reasonably treat the returned OCI image as portable and independently restorable when it is not. See [Loading snapshots](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#loading-snapshots), [Snapshotting characteristics](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#firecracker-snapshotting-characteristics), and [Where can I resume my snapshots?](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#where-can-i-resume-my-snapshots).

infrastructure/snapshot/packager.go
```diff
@@ -0,0 +1,152 @@
...
+	// The artifacts share a scratch directory; use it as the file store root.
+	workingDir := filepath.Dir(input.Artifacts[0].Path)
+
+	fileStore, err := file.New(workingDir)
+	if err != nil {
+		return nil, fmt.Errorf("creating file store: %w", err)
+	}
+	defer fileStore.Close()
+
+	// Write the microvm spec as a config blob so the image is self-describing.
```
richardcase, line 81: **[P2] The OCI image is not self-describing enough to select a compatible restore target.** Serializing the Flintlock MicroVM object does not record the Firecracker binary/version, snapshot-format compatibility, host architecture/kernel, CPU model or template, or ARM GIC version. Firecracker requires compatible snapshot formats and otherwise identical software and hardware, so a future restore path cannot safely preflight this artifact and may only discover incompatibility when loading terminates the new Firecracker process. Please include explicit VMM and host compatibility metadata in the artifact. See https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#snapshot-versioning and https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#where-can-i-resume-my-snapshots

infrastructure/microvm/firecracker/snapshot.go
```diff
@@ -0,0 +1,114 @@
...
+	memPath := scratch + "/memory"
+
+	fcClient := sdk.NewClient(vmState.SockPath(), logger, false)
+
+	pausedState := fcmodels.VMStatePaused
+	if _, err := fcClient.PatchVM(ctx, &fcmodels.VM{State: &pausedState}); err != nil {
+		return nil, fmt.Errorf("pausing microvm: %w", err)
+	}
+
+	// Always resume the VM, even if the snapshot fails, so it is never left paused.
```
richardcase, line 61: **[P1] Always resuming the source creates an undocumented clone-state security hazard.** If this artifact is later restored, the source and restored VM continue from duplicated entropy, identifiers, tokens, and application state. Firecracker explicitly classifies resuming the source and later loading its snapshot as potentially insecure. Please define and document a one-shot/migration policy, or provide semantics that can leave the source stopped when restoration is intended. See https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#snapshot-security-and-uniqueness

infrastructure/microvm/firecracker/snapshot.go
```diff
@@ -0,0 +1,114 @@
...
+) (result *ports.SnapshotResult, retErr error) {
+	logger := log.GetLogger(ctx).WithFields(logrus.Fields{
+		"service": "firecracker_microvm",
+		"vmid":    input.VMID.String(),
+	})
+	logger.Info("snapshotting microvm")
+
+	vmState := NewState(input.VMID, p.config.StateRoot, p.fs)
+
+	// Pre-flight: only a running VM can be snapshotted.
```
richardcase, line 36: **[P2] Process liveness does not establish that the guest is ready to snapshot.** This check allows a snapshot during early kernel boot, but Firecracker warns that restoring such a snapshot may crash because VMGenID interrupt handling may not yet be initialized. Gate snapshotting on guest readiness where Flintlock can observe it, or make this prerequisite explicit in the API documentation. See https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#vmgenid-device-limitation

infrastructure/microvm/cloudhypervisor/snapshot.go
```diff
@@ -0,0 +1,82 @@
...
+	scratch := vmState.SnapshotPath()
+	if err := p.fs.MkdirAll(scratch, defaults.DataDirPerm); err != nil {
+		return nil, fmt.Errorf("creating snapshot scratch dir %s: %w", scratch, err)
+	}
+
+	if err := chClient.Pause(ctx); err != nil {
+		return nil, fmt.Errorf("pausing microvm: %w", err)
+	}
+
+	// Always resume the VM, even if the snapshot fails, so it is never left paused.
```
richardcase, line 53: **[P1] Copy writable disks before resuming the source VM.** The deferred resume runs as soon as Cloud Hypervisor has written memory/config/state, but no disk images are captured before that point. The source can therefore modify its disks immediately, leaving them inconsistent with the captured memory and device state. Cloud Hypervisor explicitly warns that the disks must be copied while the source remains paused; otherwise restore can enter an undefined state. Please coordinate a disk snapshot before resuming, or leave the source paused until the caller completes one. See https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/snapshot_restore.md#preserve-the-source-vm

infrastructure/microvm/cloudhypervisor/snapshot.go
```diff
@@ -0,0 +1,82 @@
...
+	if len(destURL) >= len(fileAbsPrefix) && destURL[:len(fileAbsPrefix)] == fileAbsPrefix {
+		destURL = "file:///" + destURL[len(fileAbsPrefix):]
+	}
+	if err := chClient.Snapshot(ctx, &cloudhypervisor.VMSnapshotConfig{DestinationURL: &destURL}); err != nil {
+		return nil, fmt.Errorf("creating snapshot: %w", err)
+	}
+
+	logger.Info("snapshot taken")
+
+	return &ports.SnapshotResult{
```
richardcase, line 74: **[P1] FD-backed macvtap snapshots cannot be restored through the current client contract.** Cloud Hypervisor requires replacement descriptors through `net_fds` when restoring a VM configured with explicit network FDs, but Flintlock creates macvtap devices from raw FDs and `RestoreConfig` has no `net_fds` field. This provider nevertheless advertises snapshot support for these VMs. Please either reject snapshots for FD-backed networking until restore is supported, or persist stable network IDs and extend the restore contract to supply recreated FDs. See https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/snapshot_restore.md#restore-a-vm-with-new-net-fds

infrastructure/snapshot/packager.go
richardcase, line 81: **[P2] Record the Cloud Hypervisor version in the artifact metadata.** The client can retrieve the running VMM version through `vmm.ping`, but this config contains no provider binary/version information. A future restore path therefore cannot select or validate the Cloud Hypervisor binary against the VMM-specific `state.json`; it can only attempt a load and discover incompatibility at runtime. Please include the provider name and Cloud Hypervisor version in a versioned snapshot metadata schema.

infrastructure/microvm/cloudhypervisor/snapshot.go
```diff
@@ -0,0 +1,82 @@
...
+	cerrs "github.com/liquidmetal-dev/flintlock/core/errors"
+	"github.com/liquidmetal-dev/flintlock/core/ports"
+	"github.com/liquidmetal-dev/flintlock/pkg/cloudhypervisor"
+	"github.com/liquidmetal-dev/flintlock/pkg/defaults"
+	"github.com/liquidmetal-dev/flintlock/pkg/log"
+)
+
+const snapshotResumeTimeout = 10 * time.Second
+
+// Snapshot pauses a running microvm, captures a snapshot to disk and resumes it.
```
richardcase, line 20: **[P2] Add a provider-level test for the documented lifecycle.** The current tests cover the client request body but not this `Info → Pause → Snapshot → Resume` orchestration, the three required artifact names, file URL construction, cancellation, or resume-on-failure behavior. Since this was not runtime-verified, exercise the provider against a Unix-socket fake or add a Cloud Hypervisor integration test before advertising `SnapshotCapability`. See https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/snapshot_restore.md#snapshot-a-cloud-hypervisor-vm

richardcase (commented):
Cloud Hypervisor documentation also leaves an important public-documentation gap: callers are not told that `config.json`, `memory-ranges`, and `state.json` must be reconstructed as one snapshot directory; restored VMs begin paused; external disks and network resources must be recreated; copy-on-write restore requires the memory file to remain present and unchanged; and the available memory restore modes have host and VM-configuration constraints. Only the generated RPC schema is added here, so these restore requirements should be documented before the OCI artifact is presented as portable or self-describing. See https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/snapshot_restore.md#restore-a-cloud-hypervisor-vm

richardcase:
I'm going to relook at snapshots

PR #1173, opened 2026-09-02
feat: upgrade flintlock to containerd v2 client with hosts-dir registry auth

### LINKED ISSUE #482: Cannot pull images from a private registry
**What happened:**
I am trying to pull an image from a private registry for use as the root volume. I am using this command:

```bash
fl microvm create --host 127.0.0.1:9090 --name fctest --metadata-hostname fctest --network-interface eth1:tap  --metadata-ssh-key-file ~/.ssh/id_ed25519.pub --memory 4096 --root-image ghcr.io/richardcase/private-images/flintlock-ubuntu-perftest:7a248f6 --vcpu 2
```

The `ghcr.io/richardcase/private-images/flintlock-ubuntu-perftest:7a248f6` image is in a private package in GitHub. It requires a PAT with `repo:read` to get this.

Currently, there is no way to supply credentials and it was assumed we could supply this using the CRI section of the containerd config file like this:

```toml
  [plugins."io.containerd.grpc.v1.cri".registry]
    [plugins."io.containerd.grpc.v1.cri".registry.configs]
      [plugins."io.containerd.grpc.v1.cri".registry.configs."ghcr.io".auth]
        username = "YYYYYYYYYYY"
        password = "XXXXXXXXX"
```

However, this does not work as we aren't using the CRI.....doh!

I would expect there to be a way to supply credentials so that i can use a image from a private registry

**How to reproduce it:**
Try and use a privat eimage using the `fl` command shown above

This came via the community slack from Cory & Paul.

Paul suggested that we need to include a resolver here: https://github.com/weaveworks-liquidmetal/flintlock/blob/main/infrastructure/containerd/image_service.go#L146

**Environment:**

- flintlock version: v0.1.1
- containerd version: 1.5.9

### LINKED ISSUE #1171: Upgrade flintlock to use containerd v2
We are currently using containerd v1 internally in flintlock. We are running into issues configuring auth for private container registrys. Lets upgrade and use containerd v2 internally

So that we can take full advantage of containerd v2 features. We need to be able to pull images from private registries where the auth is configured in containerd config.

Do we need to use `WithHostsDir` when pulling images

richardcase:
## Summary
- Migrates flintlock's containerd client library from v1.7.33 to the v2.2 client (`github.com/containerd/containerd/v2`) — import path/module changes only, no behavior change to existing pull/repo/event/lease/snapshot code. Pinned to v2.2.0 rather than the newest v2.3.x release because v2.3.x requires a Go toolchain (1.26) that CI's golangci-lint isn't validated against, which caused it to OOM; v2.2.0 needs only Go 1.24.3 and exposes the same client/resolver APIs.
- Adds an optional `--containerd-hosts-dir` flag (empty/disabled by default) that lets flintlock resolve images using a `certs.d/<registry>/hosts.toml` directory, answering issue #1171's `WithHostsDir` question and giving a supported way to configure private-registry auth without editing the containerd daemon's global config. This implements the resolver-based fix proposed in #482's comment thread (`config.HostDirFromRoot` + `docker.NewResolver` + `containerd.WithResolver`), so private-registry pulls no longer require CRI.
- Regenerates the `infrastructure/mock` mocks against the v2 interfaces, updates the e2e test harness (`test/e2e/utils/runner.go`) for the v2 server config API, and documents the new flag/hosts.toml layout in `userdocs/docs/getting-started/containerd.md`.

Closes #1171.
Closes #482.

## Test plan
- [x] `go build ./...`
- [x] `go build -tags e2e ./...`
- [x] `go vet ./...` (no new issues; one pre-existing, unrelated warning in `pkg/queue`)
- [x] `go test ./...` (all packages pass, including a new unit test asserting the hosts-dir resolver option is only passed to `Pull` when configured)
- [x] CI green: build, lint, unit tests, PR title/commit message conventional-commits checks
- [ ] Manual/CI e2e run against a real containerd v2 daemon (not run in this environment)


richardcase:
Found one issue in the branch diff:

- **[P2] Correct the daemon-auth fallback guidance** (`userdocs/docs/getting-started/containerd.md:103–106`). `Client.Pull` resolves and downloads images in the client process. With `HostsDir` empty, it uses the default Docker resolver; it does not inherit the daemon’s registry credentials or mirrors. Following the current guidance leaves private-registry pulls failing even when the daemon is configured for registry access. Please correct this claim in the documentation, both flag help strings, and the associated comments to explain that client-side registry configuration must be supplied through `--containerd-hosts-dir`.

Validation: all 227 race-enabled tests passed (`go test -race ./...`), and e2e compilation passed (`go test -c -o /dev/null -tags=e2e ./test/e2e`). Live registry pulls and e2e execution were not tested.

PR #1174, opened 2026-09-02
feat: add sqlite as an option for the backing datastore

### LINKED ISSUE #1172: sqlite as an option for the backing datastore
Add sqlite as an option for storing the microvm spec definitions instead of using the containerd content store

Its easier to do queries against sqlite compared to containerd

richardcase:
## Summary
- Adds a sqlite-backed implementation of `ports.MicroVMRepository` alongside the existing containerd content-store one.
- Adds a `--repository-store` flag (`containerd` | `sqlite`, defaulting to `containerd`) and runtime factory (`infrastructure/repository`) so the backend is selectable without a compile-time swap.
- Adds an opt-in `migrate-repository` subcommand to copy existing microvm specs from containerd into sqlite ahead of a cutover.
- Extracts the containerd repo's integration test bodies into a shared `infrastructure/internal/reposcontract` suite, run against both implementations, so the two backends can't silently drift apart in behaviour.

sqlite is intended to eventually replace containerd as the backing store; this PR ships it behind a flag defaulting to the current containerd behaviour so it can be adopted incrementally.

Closes #1172

## Test plan
- [x] `go build ./...`
- [x] `go test ./...`
- [x] `golangci-lint run` on all new/changed packages
- [x] Manual smoke test: `repository.New` with the sqlite backend — save/get/list/delete/exists via a temp sqlite db, plus factory error path for an invalid `--repository-store` value
- [x] `flintlockd run --help` / `flintlockd migrate-repository --help` confirm the new flags/subcommand render correctly
- [ ] Manual end-to-end run against a real containerd instance with `migrate-repository`, for a reviewer with a test environment available


infrastructure/sqlite/repo.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	db, err := sql.Open("sqlite", cfg.DatabasePath)
+	if err != nil {
+		return nil, fmt.Errorf("opening sqlite database %s: %w", cfg.DatabasePath, err)
+	}
+
+	// sqlite only supports a single writer at a time; serialise access at the
+	// database/sql level so concurrent callers queue rather than error out.
+	db.SetMaxOpenConns(1)
```
richardcase, line 43: [P1] Coordinate SQLite access across repository instances

`serveAPI` and `runControllers` each call `inject.InitializePorts` concurrently, creating separate SQLite handles for the same database. `SetMaxOpenConns(1)` only serializes access within one pool; it does not coordinate these handles. Concurrent initialization can therefore fail with `SQLITE_BUSY` during migrations and prevent normal startup. I reproduced this by opening two repositories concurrently against each fresh database: 26 of 60 opens failed with `database is locked`.

Please share repository initialization or coordinate migrations and handle contention across connections. Add a regression test that opens multiple repositories against the same database.

infrastructure/sqlite/repo.go
```diff
@@ -0,0 +1,335 @@
...
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating microvm rows: %w", err)
+	}
+
+	return items, nil
+}
+
+// ReleaseLease is a no-op for the sqlite repository, as sqlite has no
+// equivalent concept to a containerd content-store lease.
+func (r *sqliteRepo) ReleaseLease(_ context.Context, _ *models.MicroVM) error {
```
richardcase, line 188: [P1] Preserve containerd owner-lease cleanup with SQLite

The image service still uses containerd when SQLite stores VM metadata: `PullAndMount`, `pullImage`, and `imageExists` call `withOwnerLease`. The deletion plan relies on `runtime.NewRepoRelease` to release that owner lease. Making this method a no-op removes the cleanup path, leaving leases and their resources retained after VM deletion, including for VMs migrated from the containerd repository.

Please preserve containerd lease cleanup independently of the metadata backend and cover deletion with the SQLite backend in a regression test.

richardcase (commented):
Two correctness issues remain in the SQLite backend. `make test` passed, including race detection, but the existing tests do not cover these paths.

PR #1190, opened 2026-09-04
feat: add flintlock-provision CLI to replace provision.sh

### LINKED ISSUE #603: Move `provision` script to CLI
As of now, the `provision.sh` does most of the heavy lifting to setup Flintlock and dependencies (Firecracker, containerd, networking etc) to a host machine. The idea is to port this script to a CLI.

Even though the script works great, having this feature in a CLI would make for a better developer experience, and probably easier to maintain and extend.

I believe there's still the discussion of whether this should be part of the `flintlock` CLI or maybe of a `client` (similar to what was [discussed here](https://github.com/weaveworks-liquidmetal/flintlock/discussions/592)), I ended up [creating a repo](https://github.com/lowercase00/vmctl) to start working on this, very early stages still. We can later decide where we want that to live. If anyone else also wants to work on this, we could easily just move that to a branch inside `fl` or `flintlock` though.

About the code itself, there's also the decision to use `exec.Command` vs adding more dependencies (for example, DIgitalOcean maintains a SDK to interact with `libvirtd`), just something to consider.

richardcase:
## Summary
- Adds a new `flintlock-provision` Go binary (built with cobra) that ports `hack/scripts/provision.sh`'s host-bootstrap logic — installing Firecracker, Cloud Hypervisor, containerd and flintlockd, and configuring their systemd services and devicemapper thinpool.
- Business logic lives in `internal/provision` (no CLI framework dependency, unit-testable); cobra wiring lives in `internal/command/provision`, mirroring the existing `flintlockd`/`flintlock-metrics` command structure.
- Full command parity with `provision.sh`: `all`, `apt`, `firecracker`, `cloudhypervisor`, `containerd`, `flintlock`, `direct-lvm`, `devpool`.
- Adds a `--binary-only` flag (default `false`) to the `flintlock` subcommand to install just the flintlockd binary, skipping its config file and systemd unit.
- Updates `Makefile` and `.goreleaser.yaml` with matching build/release/packaging targets for the new binary.
- `hack/scripts/provision.sh` is left untouched; deprecating it is a follow-up task.

One deliberate deviation from strict script parity: `provision.sh`'s `all`/`direct_lvm` commands have a bash parameter-expansion bug where `--thinpool` is silently ignored in favor of the `flintlock`/`flintlock-dev` constants. This port honors `--thinpool` as intended.

Closes #603

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (unit tests added for pure logic in `internal/provision`: URL/path building, arch mapping, config-file merging/rendering, thin-pool table building, network parsing)
- [x] `go run ./cmd/flintlock-provision --help` / per-subcommand `--help` reviewed for correct flag wiring
- [ ] Manual end-to-end run of each subcommand (and `all`) on a real/VM host — this provisions real host state (systemd services, LVM/devicemapper, package installs) so it needs verification outside this sandbox


internal/provision/release.go
```diff
@@ -0,0 +1,184 @@
...
+	gzr, err := gzip.NewReader(body)
+	if err != nil {
+		return fmt.Errorf("reading gzip stream from %s: %w", url, err)
+	}
+	defer gzr.Close()
+
+	tr := tar.NewReader(gzr)
+
+	for {
+		header, err := tr.Next()
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: extractTarEntry now uses a safeJoin() helper that rejects any tar entry whose resolved path would escape destDir, before it's used in any filesystem operation.

internal/provision/release.go
```diff
@@ -0,0 +1,184 @@
...
+		if err := extractTarEntry(tr, header, destDir); err != nil {
+			return err
+		}
+	}
+}
+
+func extractTarEntry(tr *tar.Reader, header *tar.Header, destDir string) error {
+	// #nosec G305 -- header.Name comes from a GitHub release tarball we control the URL of.
+	dest := filepath.Join(destDir, header.Name)
+
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: same fix as the CodeQL alert on this file - extractTarEntry now validates via safeJoin() that the entry stays within destDir before MkdirAll/OpenFile are called.

internal/provision/flintlock.go
```diff
@@ -0,0 +1,217 @@
...
+
+	parentIface := opts.ParentIface
+	if parentIface == "" && opts.BridgeName == "" {
+		iface, err := LookupInterface(runner)
+		if err != nil {
+			return err
+		}
+
+		parentIface = iface
+	}
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: AllFlintlock now resolves parentIface whenever it's empty, regardless of BridgeName - it's still needed for address auto-detection. BuildFlintlockdSettings already prefers BridgeName over ParentIface in the written config, so this doesn't change what's written.

internal/provision/network.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+
+			if addr := fields[i+1]; privateIPv4.MatchString(addr) {
+				return addr
+			}
+		}
+	}
+
+	return ""
+}
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: ParseAddressForInterface now requires an exact "dev <iface>" token match and returns immediately for an empty iface, instead of a raw substring Contains check.

internal/provision/release.go
```diff
@@ -0,0 +1,184 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return "", fmt.Errorf("building request for %s: %w", url, err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("fetching latest release for %s: %w", repo, err)
+	}
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: LatestReleaseTag now sets User-Agent and Accept: application/vnd.github+json; the shared get() helper (used for downloads/raw files) also sets User-Agent.

internal/command/provision/all.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+
+			if err := provision.InstallCloudHypervisor(cmd.Context(), runner, provision.DefaultVersion, arch); err != nil {
+				return err
+			}
+
+			containerdPaths, err := provision.AllContainerd(cmd.Context(), runner, provision.DefaultVersion, thinpool, arch, dev)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
richardcase: Fixed in a0031dc: added VersionFromEnv() plus FirecrackerVersionEnv/CloudHypervisorVersionEnv/ContainerdVersionEnv/FlintlockVersionEnv constants, and wired it into every subcommand's --version default (not just `all`) so they all honor FIRECRACKER/CLOUD_HYPERVISOR/CONTAINERD/FLINTLOCK, matching provision.sh's behaviour.

internal/provision/cloudhypervisor.go
```diff
@@ -0,0 +1,40 @@
...
+// InstallCloudHypervisor downloads and installs the given (or latest)
+// version of cloud-hypervisor for the given normalised (amd64/arm64)
+// architecture to InstallPath.
+func InstallCloudHypervisor(ctx context.Context, runner *Runner, version, normalisedArch string) error {
+	tag, err := ResolveTag(ctx, CloudHypervisorRepo, version)
+	if err != nil {
+		return fmt.Errorf("resolving cloud-hypervisor version: %w", err)
+	}
+
+	unameArch, err := FirecrackerArch(normalisedArch)
```
richardcase: This is in the cloudhypervisor command and is using a firecracker named function.  Please consider making the function generic or creating a cloudhypervisor specific one.

internal/provision/confirm.go
```diff
@@ -0,0 +1,42 @@
...
+	scanner := bufio.NewScanner(in)
+	if !scanner.Scan() {
+		return false
+	}
+
+	return strings.TrimSpace(scanner.Text()) == "y"
+}
+
+// EnsureKVM checks that /dev/kvm exists, returning an error if it doesn't.
+func EnsureKVM() error {
```
richardcase: This needs to be in a different file as it doesn't relate to "confirm"

internal/provision/containerd.go
```diff
@@ -0,0 +1,160 @@
...
+
+// StartContainerdService fetches the containerd systemd unit, points it at
+// paths.ConfigPath and starts it.
+func StartContainerdService(ctx context.Context, runner *Runner, paths ContainerdPaths) error {
+	service := ContainerdBin + ".service"
+	if err := FetchServiceFile(ctx, runner, ContainerdRepo, service, paths.ServiceFile); err != nil {
+		return err
+	}
+
+	if err := appendExecStartArg(paths.ServiceFile, "--config "+paths.ConfigPath); err != nil {
```
richardcase, line 129: [P1] `FetchServiceFile` runs `systemctl daemon-reload` before this edit, so systemd has already cached the unmodified unit. On the initial install, containerd can therefore start without `--config`; the equivalent flintlockd path can also retain `Requires=containerd.service` instead of the dev service. Please reload systemd after all unit edits and before starting it. Also consider `restart` for an already-active service, since `start` will not apply the changed configuration.

richardcase (commented):
Two actionable findings from the review:

internal/provision/lvm.go
```diff
@@ -0,0 +1,162 @@
...
+
+// ApplyLVMProfile creates (if necessary) and applies the LVM profile for thinpool.
+func ApplyLVMProfile(runner *Runner, thinpool string) error {
+	profile := filepath.Join(ThinpoolProfilePath, thinpool+"-thinpool.profile")
+
+	if err := writeFileIfMissing(profile, LVMProfile); err != nil {
+		return err
+	}
+
+	if runner.Contains(thinpool, "lvs") {
```
richardcase: [P2] This returns for every newly created pool: `AllDirectLVM` calls `CreateLogicalVolume` first, so `lvs` necessarily contains `thinpool` by the time this check runs. As a result, `lvchange --metadataprofile` is never executed and the auto-extension profile is not attached. Please make the idempotency check inspect whether the intended LV profile is already assigned, rather than whether the pool exists.

internal/provision/cloudhypervisor.go
richardcase: Fixed in 0e7d5aa: renamed FirecrackerArch to UnameArch since it's the same uname -m style mapping used by both firecracker and cloud-hypervisor, and updated both call sites.

internal/provision/confirm.go
richardcase: Fixed in 0e7d5aa: moved EnsureKVM out of confirm.go into its own kvm.go.

internal/provision/containerd.go
richardcase: Fixed in 0e7d5aa: FetchServiceFile no longer reloads systemd itself. StartContainerdService and StartFlintlockdService now call the new ReloadSystemd after their unit edits (appendExecStartArg/ReplaceRequires) and before starting, and StartService now restarts an already-active service instead of a no-op start.

internal/provision/lvm.go
richardcase: Fixed in 0e7d5aa: ApplyLVMProfile now checks via `lvs -o profile <thinpool>/thinpool` whether the intended profile is already assigned, instead of whether the volume group exists (which was always true by this point). Also noticed this same faulty check exists in provision.sh itself, so it predates the port.

Issue #1200, opened 2026-09-08
exec-api: MicroVMExec session hangs forever after guest-agent closes the vsock connection


richardcase:
## Summary

`flintlockd`'s `MicroVMExec.ExecCommand` handler (`infrastructure/grpc/exec_server.go`) hangs
indefinitely on the very first exec call against a VM, blocking any caller (e.g. a reconciler
polling readiness via a no-op `exec true`) forever. Reproduced with the official `flintlockd_amd64`
v0.14.0 release binary (current tip of `main`), Firecracker provider, `--enable-exec-api`.

## Where it hangs

`ExecCommand` opens a vsock session to the guest-agent and loops:

```go
for {
    event, err := session.Next()
    if err != nil {
        return fmt.Errorf("reading guest-agent exec session: %w", err)
    }
    resp, done := convertExecEvent(event)
    ...
    if done {
        return nil
    }
}
```

`session.Next()` (`infrastructure/vsockexec`) blocks on a socket read with **no deadline** — the
code comment on the caller side even acknowledges this:

> `session.Next()` below blocks on a socket read with no deadline; if the RPC context ends
> (client disconnect, deadline, server shutdown) close the session to unblock it instead of
> leaving the handler stuck until the guest happens to send something.

That mitigation only covers the *caller's* context ending — it doesn't cover the guest-agent
closing its end of the connection normally.

## Evidence

Firecracker's own vsock device trace log for the failing exec shows the guest-agent completing the
exchange cleanly and then closing the connection:

```
RX pkt: ... op=1 (REQUEST)                    # host opens control channel
send:   ... op=2 (RESPONSE)                   # guest accepts
RX pkt: ... len=65, op=5 (RW)                 # host writes the exec request (65 bytes)
send:   ... op=6 (CREDIT_REQUEST)
send:   ... len=5,  op=5 (RW)                 # guest-agent responds
send:   ... len=10, op=5 (RW)                 # guest-agent responds
send:   ... op=4 (SHUTDOWN), flags=3          # host shuts down both directions
RX pkt: ... op=3 (RST)                        # guest resets/closes
```

After that `RST`, `flintlockd` never logs again for that VM and the caller's RPC never returns —
no `EventExit`, no error. Confirmed this isn't a one-off: it reproduces the same way across
multiple separate VMs/runs, always immediately after the first successful exec exchange.

## Impact

Any consumer relying on `--enable-exec-api` + `MicroVMExec` to gate readiness (e.g.
[...]

PR #1201, opened 2026-09-08
fix(exec): bound guest-agent exec session reads with an idle deadline

richardcase:
## Summary

Addresses #1200: `MicroVMExec.ExecCommand` can hang forever once the
guest-agent stops sending frames on its vsock control channel, since
`Session.Next()` (`infrastructure/vsockexec/client.go`) reads with no
deadline. This bounds that blocking instead of eliminating it outright —
see "Open question on root cause" below for why I'm not claiming this is
a proven fix for the exact reported trace.

- Read frames now carry an idle deadline derived from the exec request's
  own `TimeoutSec` (already caller-controlled, no proto change needed):
  - `TimeoutSec > 0`: deadline = `TimeoutSec` + `ExecSessionIdleGrace`
    (30s) — the guest-agent already enforces `TimeoutSec` itself and is
    expected to report an outcome by then, so this only fires once the
    guest-agent should already have responded.
  - `TimeoutSec == 0` ("unbounded"): falls back to
    `ExecSessionUnboundedIdleCeiling` (15m), a last-resort circuit
    breaker rather than a liveness check, sized to not trip a real quiet
    workload while still guaranteeing the session eventually gives up on
    a wedged connection.
- `pkg/defaults`: add both constants.
- `Session.SetIdleTimeout` lets tests exercise the timeout path directly.

### Why not a flat timeout

An earlier version of this PR used a single fixed idle timeout applied
to every read. Review caught that this kills perfectly healthy
commands that are simply quiet for a while (e.g. `sleep 60`, or any
command with `timeout_seconds=120` producing no output) — the
guest-agent sends no heartbeat frames, so a quiet-but-alive command is
indistinguishable on the wire from a dead connection, and closing the
session on a false timeout kills the still-running command. Deriving
the deadline from the request's own declared budget (instead of a
constant) fixes that for bounded requests; the unbounded case still has
no fully principled answer without a guest-agent protocol change
(heartbeat frames), so it gets an explicit, generous circuit breaker
instead.

### Open question on root cause

A review comment worked through the byte counts in the issue's
Firecracker vsock trace and makes a good case that the reported 5-byte
and 10-byte guest writes match a complete `FrameExit` header +
`{"code":0}` payload — i.e. the exit frame likely *did* fully arrive,
and `Session.Next()` returns `EventExit` immediately on decoding it,
without needing EOF at all. If that's what happened, "missing EOF
detection" doesn't actually explain the reported hang, and the real
cause is still open (possibly a host-side vsock-over-UDS receive
edge-case that doesn't reduce to the framing/decoding logic touched
here). I haven't been able to reproduce the original Firecracker
scenario to confirm either way. This PR should be read as bounding the
general failure mode described in the issue (a read that can block
forever) rather than a confirmed fix for that exact trace — closing the
loop on the actual root cause needs a real repro with host-side
visibility into what the vsock UDS multiplexer actually delivers.

## Test plan

- [x] `make test` / `go build ./...`
- [x] `TestSession_Next_TimesOutWhenGuestAgentGoesQuiet` — a bounded
      session whose guest-agent goes silent without closing the
      connection or sending an exit frame times out instead of hanging
- [x] `TestSession_Next_HealthyQuietCommandDoesNotTimeOut` — a bounded
      command that's quiet for longer than the old flat timeout, but
      within its own `TimeoutSec` + grace, still completes successfully
- [x] `TestSession_Next_UnboundedTimeoutSecToleratesAQuietCommand` — an
      unbounded request tolerates an ordinary quiet interval
- [x] `TestIdleTimeoutFor` — direct unit coverage of both branches of the
      deadline derivation
- [x] Existing `TestExecServer_ExecCommand_ContextCancellationClosesSession`
      still passes unmodified

### DIFF
--- a/infrastructure/vsockexec/client.go
+++ b/infrastructure/vsockexec/client.go
@@ -6,10 +6,14 @@ package vsockexec
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/liquidmetal-dev/guest-agent/pkg/vsockclient"
+
+	"github.com/liquidmetal-dev/flintlock/pkg/defaults"
 )
 
 // EventType identifies what a Session.Next event carries.
@@ -38,7 +42,8 @@ type Event struct {
 // Session is one exec request/response exchange with a guest-agent's control
 // channel, reached by dialling a vsock UDS multiplexer path.
 type Session struct {
-	conn net.Conn
+	conn        net.Conn
+	idleTimeout time.Duration
 }
 
 // Start dials udsPath (a Firecracker/Cloud Hypervisor vsock UDS multiplexer)
@@ -57,7 +62,14 @@ func Start(ctx context.Context, udsPath string, controlPort uint32, exec *vsockc
 		return nil, fmt.Errorf("sending exec request: %w", err)
 	}
 
-	return &Session{conn: conn}, nil
+	return &Session{conn: conn, idleTimeout: defaults.ExecSessionIdleTimeout}, nil
+}
+
+// SetIdleTimeout overrides the default idle deadline Next() applies to each
+// read. Mainly useful for tests that need to exercise the timeout path
+// without waiting out the real default.
+func (s *Session) SetIdleTimeout(d time.Duration) {
+	s.idleTimeout = d
 }
 
 // SendStdin forwards p to the running command's stdin.
@@ -80,10 +92,23 @@ func (s *Session) CloseStdin() error {
 
 // Next blocks for the next frame from the guest-agent and translates it into
 // an Event. EventExit always ends the exchange.
+//
+// The underlying read carries an idle deadline: some vsock transports don't
+// reliably deliver EOF/RST to the host side when the guest-agent closes its
+// end, which would otherwise block this call forever.
 func (s *Session) Next() (Event, error) {
 	for {
+		if err := s.conn.SetReadDeadline(time.Now().Add(s.idleTimeout)); err != nil {
+			return Event{}, fmt.Errorf("setting exec session read deadline: %w", err)
+		}
+
 		f, err := vsockclient.ReadFrame(s.conn)
 		if err != nil {
+			var netErr net.Error
+			if errors.As(err, &netErr) && netErr.Timeout() {
+				return Event{}, fmt.Errorf("guest-agent exec session timed out waiting for next frame: %w", err)
+			}
+
 			return Event{}, fmt.Errorf("reading guest-agent frame: %w", err)
 		}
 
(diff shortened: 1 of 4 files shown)


infrastructure/vsockexec/client.go
```diff
@@ -80,10 +92,23 @@ func (s *Session) CloseStdin() error {
...
 // Next blocks for the next frame from the guest-agent and translates it into
 // an Event. EventExit always ends the exchange.
+//
+// The underlying read carries an idle deadline: some vsock transports don't
+// reliably deliver EOF/RST to the host side when the guest-agent closes its
+// end, which would otherwise block this call forever.
 func (s *Session) Next() (Event, error) {
 	for {
+		if err := s.conn.SetReadDeadline(time.Now().Add(s.idleTimeout)); err != nil {
+			return Event{}, fmt.Errorf("setting exec session read deadline: %w", err)
```
richardcase: [P1] Preserve healthy commands that stay quiet longer than 30 seconds

This deadline applies while the command is still running, but the guest-agent sends no heartbeat frames: a valid `sleep 60` (or a quiet build / command waiting for stdin) sends nothing until it exits. Consequently, `ExecCommand` now fails after 30 seconds even with `timeout_seconds=120` or `0` (documented as unbounded), and its deferred `session.Close()` causes the guest-agent to cancel the running process. RPC clients cannot override `SetIdleTimeout`.

Please preserve the requested command lifetime and distinguish a dead transport from a healthy quiet command (for example through a liveness mechanism), rather than imposing an unconditional output deadline. Add a regression test in which a healthy agent sends a successful exit after a quiet interval longer than the idle timeout. I reproduced this with the existing fake agent, a 100 ms idle timeout, and a successful exit scheduled at 250 ms: `Next()` fails at 100 ms instead of returning the exit event.

richardcase: Fixed — the deadline is now derived from the exec request's own `TimeoutSec` instead of a flat constant (`TimeoutSec + ExecSessionIdleGrace` when set; a 15m `ExecSessionUnboundedIdleCeiling` circuit breaker when `TimeoutSec == 0`), so it no longer imposes an unconditional deadline shorter than the command's declared budget. `SetIdleTimeout` is still there but now only overrides the derived default (used by tests to avoid waiting out the real one).

Added `TestSession_Next_HealthyQuietCommandDoesNotTimeOut` reproducing your exact repro shape (quiet interval that would trip the old flat timeout, successful exit scheduled after it) and `TestSession_Next_UnboundedTimeoutSecToleratesAQuietCommand` for the `TimeoutSec == 0` case — both now pass. `TestIdleTimeoutFor` unit-tests the derivation directly. See the PR description for the trade-off on the unbounded case (still a fixed ceiling, since there's no protocol-level liveness signal to do better without a guest-agent change).

Issue #1203, opened 2026-09-08
exec-api: replace TimeoutSec-derived idle deadline with guest-agent heartbeat once available


richardcase:
**Describe the solution you'd like:**
Once `liquidmetal-dev/guest-agent` supports periodic heartbeat frames on the exec control channel (tracking issue: [liquidmetal-dev/guest-agent#13](https://github.com/liquidmetal-dev/guest-agent/issues/13)), update `infrastructure/vsockexec.Session.Next()` to treat heartbeats as the liveness signal for its idle read deadline, replacing the `TimeoutSec`-derived deadline and `ExecSessionUnboundedIdleCeiling` fallback added in #1201 with a short, uniform timeout (e.g. 2-3x the heartbeat interval) that applies regardless of the command's declared `timeout_seconds`.

**Why do you want this feature:**
#1201 mitigates #1200 (a read that can block forever after the guest-agent goes quiet) by deriving the idle deadline from the exec request's own `TimeoutSec`, but this is a proxy for liveness, not a real signal: it can't distinguish a wedged connection from a legitimately quiet command any better than guessing at a duration. A heartbeat-based deadline would be both tighter (catching a wedged connection in seconds, not minutes) and safer (no risk of killing a healthy long-running/quiet command regardless of its declared timeout).

**Anything else you would like to add:**
Blocked on [liquidmetal-dev/guest-agent#13](https://github.com/liquidmetal-dev/guest-agent/issues/13). #1201's review discussion also raised an open question on whether the exact scenario in #1200 (a fully-delivered exit frame followed by a guest-side close) is actually explained by a missing-EOF/idle-timeout theory at all — worth revisiting once heartbeats give better visibility into what's actually happening on the wire.

richardcase:
Guest agent v0.4.0 is available

PR #1204, opened 2026-09-08
feat(exec): use guest-agent heartbeats for exec idle deadline

richardcase:
## Summary

Addresses #1203. #1201 bounded `Session.Next()`'s reads with an idle
deadline derived from the exec request's own `TimeoutSec` (plus a generous
`ExecSessionUnboundedIdleCeiling` fallback for unbounded requests), since
no real liveness signal was available at the time. That was always a
guess: it can't tell a wedged connection apart from a legitimately quiet
command.

`guest-agent` v0.4.0 now emits a `FrameHeartbeat` on a fixed 5s ticker for
the duration of a running command, independent of stdout/stderr activity
(liquidmetal-dev/guest-agent#14, fixing liquidmetal-dev/guest-agent#13).
This PR uses that as the real liveness signal:

- Bump `github.com/liquidmetal-dev/guest-agent` to v0.4.0.
- `pkg/defaults`: replace `ExecSessionIdleGrace` /
  `ExecSessionUnboundedIdleCeiling` with a single
  `ExecSessionIdleTimeout` (15s, 3x the heartbeat interval).
- `infrastructure/vsockexec`: `Session` now applies that flat timeout to
  every read regardless of the request's `TimeoutSec` (still passed
  through to the guest-agent unchanged, which continues to enforce it
  itself). `Next()` consumes `FrameHeartbeat` to re-arm the deadline
  without surfacing it as an `Event` — callers still only ever see
  stdout/stderr/exit/error events.

## Test plan

- [x] `make test` / `go build ./...` / `go vet ./...`
- [x] `TestSession_Next_TimesOutWhenGuestAgentGoesQuiet` — a session that
      gets no further frames (no heartbeat, no exit) still times out
      instead of hanging
- [x] `TestSession_Next_HeartbeatsKeepQuietSessionAlive` — a quiet command
      bridged entirely by heartbeats survives a shrunk idle timeout, for
      both a bounded and an unbounded `TimeoutSec`
- [x] `TestSession_Next_HeartbeatNotSurfacedAsEvent` — heartbeat frames are
      swallowed internally, never returned as an `Event`
- [x] Existing `TestSession_ExecStdoutStderrExit` and
      `TestSession_SendStdinAndCloseStdin` pass unmodified

Note: local `golangci-lint` (v1.63.4, built with go1.25) can't typecheck
against this repo's `go 1.26` directive and fails the same way on
unrelated third-party packages — likely a local toolchain mismatch, not
something introduced here. Worth confirming CI's golangci-lint version
handles `go 1.26` cleanly.


infrastructure/vsockexec/client.go
```diff
@@ -86,12 +65,12 @@ func Start(ctx context.Context, udsPath string, controlPort uint32, exec *vsockc
 		return nil, fmt.Errorf("sending exec request: %w", err)
 	}
 
-	return &Session{conn: conn, idleTimeout: idleTimeoutFor(exec.TimeoutSec)}, nil
+	return &Session{conn: conn, idleTimeout: defaults.ExecSessionIdleTimeout}, nil
```
richardcase: [P1] Gate the short deadline on guest heartbeat support

This applies the 15s deadline to existing VMs running guest-agent v0.3.0 too. Updating `go.mod` only updates the host client; it does not replace the binary inside those VMs. Both agent versions still speak protocol version 1, so the exec request is accepted, but v0.3.0 never emits heartbeats. Consequently, a healthy `sleep 16` with `TimeoutSec: 30` now fails after 15s; the gRPC handler then closes the session, which the agent treats as cancellation. I reproduced the timeout at 15.007s with a fake legacy guest scheduled to send a successful exit at 16s, without overriding `SetIdleTimeout`.

Please retain the previous deadline until heartbeat support is established (via capability/version detection or an observed heartbeat), or explicitly reject unsupported agents before starting the command. Add a legacy-agent regression test so a host upgrade cannot silently turn ordinary quiet commands into timeouts.

richardcase (commented):
Reviewed the heartbeat handling, guest-agent protocol compatibility, and gRPC session lifecycle. `make test` passed, as did focused race tests for `infrastructure/vsockexec` and `infrastructure/grpc`. One compatibility regression is noted inline; reproduced with a fake legacy guest using the unchanged production idle timeout.

infrastructure/vsockexec/client.go
richardcase: Good catch, fixed in 2347b1c: the session now starts on the legacy TimeoutSec-derived deadline (unchanged from #1201) and only switches to the tight heartbeat deadline once a FrameHeartbeat is actually observed on that session. A pre-v0.4.0 guest-agent that never sends one keeps the TimeoutSec-derived deadline for the life of the session.

Added `TestSession_Next_LegacyAgentWithoutHeartbeatsUsesTimeoutSecDeadline` as the regression test, plus `TestSession_Next_HeartbeatSwitchesToShortDeadline` to prove the switch actually takes effect once a heartbeat does arrive.

PR #1223, opened 2026-09-16
fix(userdocs): force transitive npm deps to patched versions

richardcase:
## Summary
- Adds a `resolutions` block to `userdocs/package.json` forcing 11 vulnerable transitive npm dependencies to their patched versions: `brace-expansion`, `image-size`, `json5`, `minimatch`, `on-headers`, `postcss`, `postcss-selector-parser`, `qs`, `uuid`, `ws`, `body-parser`.
- Resolves 33 open Dependabot security alerts. These are all transitive dependencies of the docusaurus/webpack toolchain (not direct deps in `package.json`), and several have multiple vulnerable version ranges coexisting side by side in the tree (e.g. both a 1.x and a 2.x line of `brace-expansion`), so a plain `yarn upgrade <pkg>@<version>` can't reach every instance — Yarn's `resolutions` field is the mechanism for forcing a single version regardless of what range any consumer declares.

**Deliberately not touched, with reasons:**
- `serialize-javascript` — every patched 7.x release requires Node >=20, but the Netlify deploy environment for flintlock-docs builds with Node 18, so this resolution broke the live docs deploy (`The engine "node" is incompatible... Expected version ">=20.0.0". Got "18.20.8"`). Our GitHub Actions build didn't catch this since it runs on Node 22. Fixing these 2 alerts needs a Node-version bump on the Netlify side first, which is a separate infra decision.
- `js-yaml` — forcing it to the patched 4.x line breaks the build: `gray-matter` (used internally by Docusaurus for markdown frontmatter parsing) calls the removed `yaml.safeLoad`, so `yarn build` fails immediately with `Function yaml.safeLoad is removed in js-yaml 4`. Fixing this needs the two vulnerable version families (3.x and 4.x) resolved independently, which classic Yarn resolutions can't express without scoping to the exact dependent package paths.
- `ajv` and `path-to-regexp` — both have genuinely incompatible major versions coexisting in the tree for different consumers (e.g. Express intentionally pins an old `path-to-regexp` due to past ReDoS fixes); forcing a single version tree-wide risks silently breaking whichever consumer needs the older API.
- `webpack-dev-server` — the patched version is a major bump (4→5) and this package is dev-server-only, never exercised by `yarn build` or this repo's CI, so there's no way to verify a bump doesn't break local `yarn start` for docs contributors.

These remaining alerts are left for Dependabot's own PRs, which have the actual dependency-graph awareness needed to scope fixes safely per-consumer.

## Test plan
- [x] `cd userdocs && yarn install && yarn build` succeeds on Node 22 (this repo's default) and Node 18.20.8 (matching Netlify's deploy environment)
- [x] CI green (lint/test/Build docusaurus)


richardcase:
@dependabot rebase

richardcase:
@dependabot recreate

